PR #5, opened 2017-02-10
Replace use of wp_get_http()

lukecav:


lukecav:
Sorry for the lack of wording on the commits, my mistake.

wordpress-importer.php
```diff
@@ -988,7 +988,19 @@ function fetch_remote_file( $url, $post ) {
 			return new WP_Error( 'upload_dir_error', $upload['error'] );
 
 		// fetch the remote url and write it to the placeholder file
-		$headers = wp_get_http( $url, $upload['file'] );
+		$response = wp_remote_get( $url, array( 'timeout' => 60) );
```
ocean90: Missing space after 60, but why is there even a timeout?

wordpress-importer.php
```diff
@@ -988,7 +988,19 @@ function fetch_remote_file( $url, $post ) {
...
+		$headers = wp_remote_retrieve_headers( $response );
+		if ( is_wp_error() | '200' !=  wp_remote_retrieve_response_code( $response ) ) {
```
ocean90: The argument for `is_wp_error()` is missing. Should also use a strict check there.

wordpress-importer.php
ocean90: Missing space after 60, but why is there even a timeout?

wordpress-importer.php
```diff
@@ -988,7 +988,19 @@ function fetch_remote_file( $url, $post ) {
...
+    			return $headers;
+		}
+
+		$out_fp = fopen( $file_path, 'w' );
```
boonebgorges: `$file_path` is not defined

lukecav:
I think I got those both fixed, please take a look.

raversnet:
I keep getting:

Parse error: syntax error, unexpected 'if' (T_IF) on Line 993 with the changes

lukecav:
@raversnet 

Not had that issue, when I have tested the plugin.

raversnet:
Should have posted back.  Got it working and works great.  Thanks

lukecav:
@mcbsys 

Will make the change now.

ocean90:
@lukecav As a WordPress developer, at least that's what your profile says, you should be aware of the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
Instead of editing the files on GitHub directly you should use a local checkout so you can test what you're changing here. For example you're still using a bitwise operator.

Also, as already mentioned, this PR should only include changes to `wp_get_http()`. I suggest to create a new one which only contains the required changes to remove the use of `wp_get_http()`, preferably with better commit messages.

lukecav:
I did test the changes made on local dev. I do not see the issues with making the charges on my fork of the WordPress Importer. But I will create a separate pull request for the other fix, my mistake.

ariona:
The importer works with no warning/error, but on the media side, it only import metadata but not the files. any idea?

kaiserkiwi:
Tried this fix, but only get empty entrys. The upload works, the files are on my server. The media library shows all entrys, but they are empty. The link to the post is empty, the url is nearly empty. Only the image name is used without the file extension. No upload path, nothing. 

Instead of _https://domain.com/wp-content/uploads/2017/04/filename.jpg_ the path is only _https://domain.com/filename/_ and this goes nowhere. 

So the upload clearly works, but the entrys in the database are completely wrong.

PR #6, opened 2017-03-30
Add wp_slash before importing comment

mcbsys:
Backslashes are removed from comments on import. Not sure if this is the right place to add wp_slash?

See https://core.trac.wordpress.org/ticket/21007#comment:10

### DIFF
--- a/wordpress-importer.php
+++ b/wordpress-importer.php
@@ -787,6 +787,7 @@ class WP_Import extends WP_Importer {
 						if ( isset( $inserted_comments[$comment['comment_parent']] ) )
 							$comment['comment_parent'] = $inserted_comments[$comment['comment_parent']];
 						$comment = wp_filter_comment( $comment );
+						$comment = wp_slash( $comment );
 						$inserted_comments[$key] = wp_insert_comment( $comment );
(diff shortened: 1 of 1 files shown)


wordpress-importer.php
```diff
@@ -787,6 +787,7 @@ function process_posts() {
 						if ( isset( $inserted_comments[$comment['comment_parent']] ) )
 							$comment['comment_parent'] = $inserted_comments[$comment['comment_parent']];
 						$comment = wp_filter_comment( $comment );
+						$comment = wp_slash( $comment );
```
ocean90: The data should get slashed before calling `wp_filter_comment()`. This would make it consistent with current behaviour in core:

* https://github.com/WordPress/WordPress/blob/e357195ce303017d517aff944644a7a1232926f7/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php#L593
* https://github.com/WordPress/WordPress/blob/76f44d242bb87ed47281aeb0750b39c8a4b00571/wp-includes/comment.php#L3095

mcbsys: @ocean90 thanks for reviewing. I think I've fixed that, moving the wp_slash up one line.

PR #10, opened 2017-05-12
Fix issue with wp_get_http()

gerasimovdaniel:
Replace wp_get_http() with more appropriate wp_safe_remote_get().


leek:
**@gerasimovdaniel** I believe this was already fixed in #5. You can safely close this PR.

PR #11, opened 2017-08-03
Adding composer.json

leek:
Simply adding a `composer.json` file so this plugin can be installed via Composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,14 @@
+{
+    "name": "boonebgorges/wordpress-importer",
+    "description": "WordPress Importer",
+    "type": "wordpress-plugin",
+    "license": "GPLv2+",
+    "keywords": [
+        "wordpress",
+        "plugin"
+    ],
+    "require": {
+        "php": ">=5.3.0",
+        "composer/installers": "~1.0"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "boonebgorges/wordpress-importer",
```
GaryJones: The owner needs updating.

composer.json
```diff
@@ -0,0 +1,14 @@
...
+    "description": "WordPress Importer",
+    "type": "wordpress-plugin",
+    "license": "GPLv2+",
```
GaryJones: The license should now be `"GPLv2-or-later"`

composer.json
```diff
@@ -0,0 +1,14 @@
...
+    "keywords": [
+        "wordpress",
+        "plugin"
+    ],
+    "require": {
+        "php": ">=5.3.0",
```
GaryJones: While Composer needs 5.3, what code in this plugin makes it 5.3?

ocean90:
Closing for now. A proper file should be added as part of #30.

Issue #18, opened 2017-12-15
Regex parser: Single line `<wp:term>` elements no longer work


dd32:
During the process of attempting to tag a 0.6.4 release, I found that the fix in #3 for #2 no longer handles single-line elements.

For example, from the core unit test's valid-wxr-1.0.xml:
`<wp:tag><wp:tag_slug>face</wp:tag_slug><wp:tag_name><![CDATA[face]]></wp:tag_name></wp:tag>`

It also doesn't support things with the following formats:
`<wp:tag>\n <..><..></wp:tag>`
`<wp:tag><..><..>\n</wp:tag>`
`<wp:tag><..>\n<..></wp:tag>`

I'll submit a PR with test coverage of this in a bit, after #15 is fixed.

dd32:
Fixed by #21

PR #21, opened 2017-12-15
Fix support for single-line terms

dd32:
This fixes support for single-line `<wp:tag>`, `<wp:term>` and `<wp:category>` tags in the regex parser and also allows for them to be broken over multiple lines.


dd32:
Closes #18

PR #23, opened 2017-12-21
Add new hooks to allow additional processing

jfarthing84:
Adds two new hooks, `before_process_import` and `after_process_import`, to make it more intuitive to add additional processing to the import process.

### DIFF
--- a/src/wordpress-importer.php
+++ b/src/wordpress-importer.php
@@ -106,7 +106,9 @@ class WP_Import extends WP_Importer {
 		$this->import_start( $file );
 
 		$this->get_author_mapping();
-
+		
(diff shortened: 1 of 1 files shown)


jrfnl:
@jfarthing84 It might be helpful to add docblocks for each of the new hooks ?

dd32:
Should some kind of context/extra information be provided to these hooks to make them more useful?

jfarthing84:
No, not really.

dd32:
Not even `$this`?

jfarthing84:
The other actions called in this plugin have no arguments, why should these? It wouldn't hurt, of course.

PR #31, opened 2018-03-03
CS: Fix all code standards violations

GaryJones:
- Adds supporting files such that the latest releases of PHPCS, WPCS, PHPCompatibility (and PHPUnit 5.7) are available and configured locally.
- Addresses all of the violations.

screenshot 2018-03-03 14 17 40

All of the (limited number of) unit tests still pass.

screenshot 2018-03-03 14 18 11


phpunit/bootstrap.php
```diff
@@ -5,10 +5,12 @@
  * @package Sample_Plugin
  */
 
+// phpcs:disable
```
jrfnl: Just wondering why you are using an `// phpcs:disable` here.

Why not use an `<exclude>` from the ruleset if you are disabling PHPCS for this file for all sniffs ? That way, the file won't even get tokenized.
Or you could use a ` // @phpcs:ignoreFile` in the file docblock ?
Or if you did it this way because you do want the file docblock checked, but nothing else, you could put the `phpcs:disable` at the end of the docblock to disable PHPCS for the rest of the file.

.phpcs.xml.dist
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress IMporter">
```
jrfnl: `IMporter` => `Importer`

.phpcs.xml.dist
```diff
@@ -0,0 +1,51 @@
...
+	<exclude-pattern>vendor/</exclude-pattern>
+
+	<!-- How to scan -->
+	<arg value="sp"/> <!-- Show sniff and progress -->
+	<arg name="basepath" value="."/><!-- Strip the file paths down to the relevant bit -->
+	<arg name="colors"/>
+	<arg name="extensions" value="php"/>
+	<arg name="report" value="full"/>
+	<arg name="report" value="summary"/>
+	<arg name="report" value="source"/>
```
jrfnl: Generally speaking, `report` is one of those arguments you would normally not put in the ruleset to allow devs more flexibility.

If you do wish to keep it there, you can merge the three commands (PHPCS 3.x):
```xml
<arg name="report" value="full,summary,source"/>
```

.phpcs.xml.dist
```diff
@@ -0,0 +1,51 @@
...
+
+	<!-- Rules: Check PHP version compatibility -->
+	<config name="testVersion" value="5.2-"/>
+	<rule ref="PHPCompatibility"/>
+
+	<!-- Rules: WordPress Coding Standards -->
+	<config name="minimum_supported_wp_version" value="3.6"/>
+	<rule ref="WordPress">
+		<exclude name="WordPress-VIP"/>
```
jrfnl: Did you mean to do `<exclude name="WordPress-VIP"/>` ? Or did you intend it as `<exclude name="WordPress.VIP"/>` ?

phpunit/tests/import.php
```diff
@@ -20,29 +34,37 @@ function setUp() {
 		add_filter( 'import_allow_create_users', '__return_true' );
 
 		global $wpdb;
-		// crude but effective: make sure there's no residual data in the main tables
+		// Crude but effective: make sure there's no residual data in the main tables.
 		foreach ( array( 'posts', 'postmeta', 'comments', 'terms', 'term_taxonomy', 'term_relationships', 'users', 'usermeta' ) as $table ) {
-			$wpdb->query( "DELETE FROM {$wpdb->$table}" );
+			$wpdb->query( "DELETE FROM {$wpdb->$table}" ); // phpcs:ignore
```
jrfnl: Should this ignore be made more specific ? I.e. just ignore one or more sniffs ? or a sniff category ?
Now it's a blanket ignore for everything and will also not report on new issues being introduced into the line.

phpunit/tests/base.php
```diff
@@ -1,5 +1,17 @@
 <?php
+/**
+ * WordPress Importer plugin testcase.
+ *
+ * @package WordPress
```
jrfnl: If two tags directly follow each other, shouldn't their values be aligned with each other ?

```php
/**
 * @package    WordPress
 * @subpackage Importer
 */
```

phpunit/tests/import.php
```diff
@@ -76,7 +98,7 @@ function test_small_import() {
 		$comment_count = wp_count_comments();
 		$this->assertEquals( 1, $comment_count->total_comments );
 
-		$posts = get_posts(
+		$posts = get_posts( // phpcs:ignore WordPress.VIP.RestrictedFunctions.get_posts_get_posts
```
jrfnl: I've seen this ignore a couple of times in the test files - why not just exclude the sniff for the test directory from the ruleset ?

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+		}
+	}
+
+	/**
+	 * Create new categories based on import information.
+	 *
+	 * Doesn't create a new category if its slug already exists.
+	 */
+	public function process_categories() {
+		$this->categories = (array) apply_filters( 'wp_import_categories', $this->categories );
```
jrfnl: Looks like the `(array)` is a new introduction.

* Should this be in a CS PR ?
* Should the array cast be done before or after the `empty()` check ? or should there maybe not be an array cast at all and the `empty()` check extended with an `is_array()` check ?

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+		unset( $this->categories );
+	}
+
+	/**
+	 * Create new post tags based on import information.
+	 *
+	 * Doesn't create a tag if its slug already exists.
+	 */
+	public function process_tags() {
+		$this->tags = (array) apply_filters( 'wp_import_tags', $this->tags );
```
jrfnl: Same here - see https://github.com/WordPress/wordpress-importer/pull/31/files#r172020244

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+				'description' => $tag_desc,
+			);
+
+			$id = wp_insert_term( $tag['tag_name'], 'post_tag', $tagarr );
+			if ( ! is_wp_error( $id ) ) {
+				if ( isset( $tag['term_id'] ) ) {
+					$this->processed_terms[ intval( $tag['term_id'] ) ] = $id['term_id'];
+				}
+			} else {
+				printf( __( 'Failed to import post tag %s', 'wordpress-importer' ), esc_html( $tag['tag_name'] ) );
```
jrfnl: Looks like a translators comment is needed here.

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+				'parent'      => intval( $parent ),
+			);
+
+			$id = wp_insert_term( $term['term_name'], $term['term_taxonomy'], $termarr );
+			if ( ! is_wp_error( $id ) ) {
+				if ( isset( $term['term_id'] ) ) {
+					$this->processed_terms[ intval( $term['term_id'] ) ] = $id['term_id'];
+				}
+			} else {
+				printf( __( 'Failed to import %1$s %2$s', 'wordpress-importer' ), esc_html( $term['term_taxonomy'] ), esc_html( $term['term_name'] ) );
```
jrfnl: Looks like a translators comment is needed here.

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+		/**
+		 * Filters the metadata attached to an imported term.
+		 *
+		 * @since 0.6.2
+		 *
+		 * @param array $termmeta Array of term meta.
+		 * @param int   $term_id  ID of the newly created term.
+		 * @param array $term     Term data from the WXR import.
+		 */
+		$term['termmeta'] = (array) apply_filters( 'wp_import_term_meta', $term['termmeta'], $term_id, $term );
```
jrfnl: Same again - see https://github.com/WordPress/wordpress-importer/pull/31#discussion_r172020244

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+	/**
+	 * Create new posts based on import information.
+	 *
+	 * Posts marked as having a parent which doesn't exist will become top level items.
+	 * Doesn't create a new post if: the post type doesn't exist, the given post ID
+	 * is already noted as imported or a post with the same title and date already exists.
+	 * Note that new/updated terms, comments and meta are imported for the last of the above.
+	 */
+	public function process_posts() {
+		$this->posts = (array) apply_filters( 'wp_import_posts', $this->posts );
```
jrfnl: And again - see https://github.com/WordPress/wordpress-importer/pull/31#discussion_r172020244

src/class-wp-import.php
```diff
@@ -0,0 +1,1420 @@
...
+
+		foreach ( $this->posts as $post ) {
+			$post = apply_filters( 'wp_import_post_data_raw', $post );
+
+			if ( ! post_type_exists( $post['post_type'] ) ) {
+				printf(
+					__( 'Failed to import &#8220;%1$s&#8221;: Invalid post type %2$s', 'wordpress-importer' ),
```
jrfnl: Missing translators comment

src/wordpress-importer.php
```diff
@@ -1,1236 +1,65 @@
...
-	if ( file_exists( $class_wp_importer ) )
-		require $class_wp_importer;
-}
-
-// include WXR file parsers
-require dirname( __FILE__ ) . '/parsers.php';
-
 /**
- * WordPress Importer class for managing the import process of a WXR file
+ * WordPress Importer Git loader
```
jrfnl: One more - looks like this file description got incorrectly moved from the root `wordpress-importer.php` file to this one.

phpunit/tests/base.php
GaryJones: Because currently WP Core doesn't align that:

phpunit/tests/import.php
GaryJones: It was a quick way to ignore multiple issues. More specific fixings could be done once the mass of issues are done.

.phpcs.xml.dist
GaryJones: The latter - tried to add it back in manually.

phpunit/bootstrap.php
GaryJones: There was too many issues to address in a suitable format - I didn't want to touch too much.

`// phpcs:disable` made most sense to me at the time, but I can update this.

GaryJones:
Thanks for the review - I'll get those addressed.

phpunit/tests/base.php
jrfnl: Hmm.. surprising.

composer.json
```diff
@@ -1,9 +1,9 @@
...
     "require-dev": {
-        "squizlabs/php_codesniffer": "^3.2",
-        "wp-coding-standards/wpcs": "^0.14.1",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
-        "wimg/php-compatibility": "^8.1",
-        "phpunit/phpunit": "^5.7"
+        "phpcompatibility/phpcompatibility-wp": "^1.0",
+        "phpunit/phpunit": "^5.7",
+        "squizlabs/php_codesniffer": "^3.3",
+        "wp-coding-standards/wpcs": "^0.14"
```
jrfnl: Isn't this suppose to be `^1.0.0` ?

phpunit/tests/import.php
jrfnl: This `ignore` can now be removed as VIP is no longer included. You may want to do a search against `phpcs:` to see if there are more ignores which can be removed.

phpunit/tests/import.php
```diff
@@ -260,25 +288,27 @@ function test_ordering_of_importers() {
...
+		$this->import_wp( DIR_TESTDATA_WP_IMPORTER . '/slashes.xml', $authors );
 
 		$alpha = get_term_by( 'slug', 'alpha', 'category' );
 		$this->assertSame( 'a \"great\" category', $alpha->name );
 
 		$tag1 = get_term_by( 'slug', 'tag1', 'post_tag' );
 		$this->assertSame( "foo\'bar", $tag1->name );
 
-		$posts = get_posts(
+		$posts = get_posts( // phpcs:ignore WordPress.VIP.RestrictedFunctions.get_posts_get_posts
```
jrfnl: Here's another ignore which can be removed.

composer.json
GaryJones: er....yup.

GaryJones:
All of @jrfnl's feedback addressed.

The only outstanding reported violations are:

```
456 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.NoNonceVerification)
 462 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.NoNonceVerification)
 467 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.NoNonceVerification)
 468 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.NoNonceVerification)
 476 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.NoNonceVerification)
 477 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.NoNonceVerification)
```

src/class-wp-import.php
```diff
@@ -521,9 +521,9 @@ public function get_author_mapping() {
 	 * Doesn't create a new category if its slug already exists.
 	 */
 	public function process_categories() {
-		$this->categories = (array) apply_filters( 'wp_import_categories', $this->categories );
+		$this->categories = apply_filters( 'wp_import_categories', $this->categories );
 
-		if ( empty( $this->categories ) ) {
+		if ( ! is_array( $this->categories ) || empty( $this->categories ) ) {
```
jrfnl: 1. My previous remark was that this kind of logic change doesn't belong in this PR. Changing it from a cast to an `is_array()` check, doesn't address that.
2. The `is_array()` and `empty()` need to be reversed as you are now testing the variable for being an array before you've tested whether it is actually set. In the above, that's not an issue, but in at least one of the other places it is nd it's the reason the unit tests are failing for this PR.
```php
if ( empty( $this->categories ) || ! is_array( $this->categories ) ) {
```

src/class-wp-import.php
```diff
@@ -708,9 +708,9 @@ protected function process_termmeta( $term, $term_id ) {
 		 * @param int   $term_id  ID of the newly created term.
 		 * @param array $term     Term data from the WXR import.
 		 */
-		$term['termmeta'] = (array) apply_filters( 'wp_import_term_meta', $term['termmeta'], $term_id, $term );
+		$term['termmeta'] = apply_filters( 'wp_import_term_meta', $term['termmeta'], $term_id, $term );
 
-		if ( empty( $term['termmeta'] ) ) {
+		if ( ! is_array( $term['meta'] ) || empty( $term['termmeta'] ) ) {
```
jrfnl: :point_up: This is the one which causes the unit tests to fail. Then again, you may have intended the `is_array()` to check `$term['termmeta']`, not, as it does now, `$term['meta']`.

GaryJones:
Array casts and checks removed.

I've rewritten the mess of commits into three simpler commits - one adds the extra project files and configs, one splits classes into their own files, one fixes most of the violations.

GaryJones:
@ocean90 @dd32 Who can I nudge to get this looked at?

ocean90:
Unfortunately, this PR needs to be redone. But instead of fixing everything in one step it should be broken up into multiple PRs. Moving classes into their own files was already done in #68.

szepeviktor:
@ocean90 @GaryJones Please get me a reviewer over at #31

PR #37, opened 2018-05-27
Added base_blog_url field

palmiak:
I added the `base_blog_url` variable,

This is needed to fix importing multiple files bug in WP-CLI https://github.com/wp-cli/import-command/pull/27

The problem is that WP-importer files from WPMU have the same base url for every blog, to recognise their  url the `base_blog_url` is needed. Having this we will able to group the processed post IDs correctly.


schlessera:
Gentle post-Gutenberg ping to see if PRs like the above can get reviewed/merged...

palmiak:
@ocean90 Thanks :)

PR #39, opened 2018-06-23
Add ability to prevent remote fetches

edavis:
Hello — Figured I'd open this PR to see if there was any interest. If not, that's okay.

The use case is actually a real-world scenario I'm dealing with right now :)

I've got a site with over 2,000 attachments I'm trying to import into an existing site. To speed things up, I downloaded all the files inside the uploads directory to my local machine via FTP.

I then created an export on the source site (All content + Media) and then started an import on the destination site. Problem is, the destination site still wants to download all the files from the source site even though they already exist locally.

With this PR, a new constant for wp-config.php is introduced (`IMPORT_PREVENT_REMOTE_FETCH`) and when set to true the `fetch_remote_file()` function is short-circuited to return an array of info on the locally existing file for the attachment. There is also a `import_prevent_remote_fetch` filter that receives the value of the constant, the URL, and the attachment to provide more fine-tuned control.

Looking forward to any feedback. Thanks!

### DIFF
--- a/src/wordpress-importer.php
+++ b/src/wordpress-importer.php
@@ -991,6 +991,28 @@ class WP_Import extends WP_Importer {
 		// extract the file name and extension from the url
 		$file_name = basename( $url );
 
+		// don't fetch files, if requested
+		$prevent = defined( 'IMPORT_PREVENT_REMOTE_FETCH' ) && IMPORT_PREVENT_REMOTE_FETCH;
+		if ( apply_filters( 'import_prevent_remote_fetch', $prevent, $url, $post )) {
+			printf( __( "Skipping fetch of '%s'\n", 'wordpress-importer' ), $url );
+			$upload_dir = wp_upload_dir( $post['upload_date'] );
+			$attachment_url = sprintf( '%s/%s', $upload_dir['url'], $file_name );
+			$attachment_filename = sprintf( '%s/%s', $upload_dir['path'], $file_name );
+			$attachment_type = wp_check_filetype( $attachment_filename );
+			$upload = array(
+				'url' => $attachment_url,
+				'file' => $attachment_filename,
+				'type' => $attachment_type['type'],
+				'error' => false,
+			);
+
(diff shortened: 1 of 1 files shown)


ChrisHardie:
Just wanted to say thanks for creating this - it came in very handy while creating the Flickr-to-WordPress migration tool at https://github.com/ChrisHardie/flickr-to-wordpress (described in some more detail at https://chrishardie.com/2019/02/moving-photos-flickr-wordpress/ ). I appreciate it!

edavis:
@ChrisHardie You're very welcome! Glad you found it useful. Your photos site looks great!

steevhise:
hmm looks like this patch no longer works...

steevhise:
looks like the same functionality is now in place for the plugin, the import_allow_fetch_attachments filter. maybe.

dd32:
> looks like the same functionality is now in place for the plugin, the import_allow_fetch_attachments filter. maybe.

The functionality of that filter is that it skips importing attachments entirely.

This PR is with the intention of skipping the remote-fetch of attachments, for when the attachment already exists on disk but not within the media library. So the file isn't created, but the WP_Post is.  
_I don't personally feel like this is a use-case that would be added to the plugin myself._

carstingaxion:
> This PR is with the intention of skipping the remote-fetch of attachments, for when the attachment already exists on disk but not within the media library. So the file isn't created, but the WP_Post is.

_THIS_ is exactly my use-case. 

I'm would like to use WordPress Playground as a tool to not only show, but also edit & save demo-data. So I created a blueprint, that imports all data from a wxr inside a Github repo. After editing your imported content in Playground, it is exported as wxr and commited back to the repo. _(That's the short version, of what's happening)_ 

New uploaded files, get also added to the repo with the final commit.

But next time the Playground is started the import starts again, and that is the moment, where it would be absolutely use- and helpful, if we could rely on:

> the attachment already exists on disk [...] So the file isn't created, but the WP_Post is.

PR #45, opened 2018-12-09
Source the WP_Import class regardless of request context

jblz:
This change short-circuits the instantiation instead of the declaration of the `WP_Import` class (& other backing resources).

This is instead of preventing the file from being sourced altogether when the `WP_LOAD_IMPORTERS` constant is not defined.

As is, it's difficult (if not impossible) to modify that constant in a plugin prior to its evaluation here.

---

I inspected what runs when the file is sourced and there doesn't seem to be anything that's resource intensive or that would introduce side-effects.


ocean90:
I'm going to close this as wontfix since it's now possible to include the `WP_Import` class separately, see #68.

PR #49, opened 2019-02-23
Move build bundle to `build/`

spen:
This PR moves the build target to `build/` and ignores that directory so that we're not checking in built changes on each update.

### Testing

- Checkout this branch
- Make sure there's no `src/build.js` file
- Run `npm run build`
- Make sure the UI works just as before


spen:
Eek, wrong base repository, sorry y'all!

PR #52, opened 2019-07-15
Update Travis script & minimum WP requirement

jrfnl:
* Don't use `sudo`. This is [no longer supported on Travis](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).
* Cache the correct directories for Composer.
* Add testing against PHP 7.3 and PHP 7.4 (dev).
* Simplify the Xdebug removal command.
* Fix the test matrix.
    WP `trunk/latest` no longer support PHP < 5.6.20.
    The plugin claims to support WP 3.6 and higher, however this can no longer be tested as due to a change in the unit test setup of WP Core in WP 3.7, the unit tests can not be run in combination with WP 3.6 anymore.
    So for the highest stable PHP version, add a build against the lowest testable WP version, 3.7 and update the minimum WP requirement to WP 3.7.
    For the lowest supported PHP version, `5.2`, a build against WP 3.7 fails with warnings about incorrect use of `wpdb::prepare()`, so, for now, test this against a low version which doesn't fail on those warnings.
    And for all other PHP versions below 5.6.20, fix the WP version being tested against at a version below WP 5.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
-sudo: false
 dist: trusty
 language: php
 
 cache:
   directories:
     - vendor
-    - $HOME/.composer/cache
+    # Cache directory for older Composer versions.
+    - $HOME/.composer/cache/files
+    # Cache directory for more recent Composer versions.
+    - $HOME/.cache/composer/files
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


jrfnl:
The build failure on PHP 7.4 will be fixed by PR #53.

Issue #55, opened 2019-10-06
.bypostauthor class is not added at comments


DesignerThan95:
At the "Template: Comments" Post, the `.bypostauthor` class is not added to the comments, because they don't have the right `user_id` set.

DesignerThan95:
I have added code that searches for the user name of the comment and if one is found, the id gets added to the comment. Through that change, the `.bypostauthor` Class gets added to the comments.

DesignerThan95:
As I said in my pull request I tested the Import again and if the user_id will be set right.
The data I import is the following: https://github.com/WPTRT/theme-unit-test
I do import the users of the test data and dont take an existing user and dont create a new one.
And if I do that the `.bypostauthor` class is not added because the user_id of the comment is zero. Do you guys have a tip for me where I should look for a solution?

Then I have tried it with creating new users and I registered that although the posts are transferred to the comments werent. Is that thought so or would it be good to solve that? If there should be a solution, I could take some time and learn how to solve it if you guys give me a tip where I should start looking.

PR #56, opened 2019-10-06
bugfix: user_id is now added to the comment if the user is found (#55)

DesignerThan95:
Fixed the issue with the `.bypostauthor` class not added to the comments

Issue #55

### DIFF
--- a/src/wordpress-importer.php
+++ b/src/wordpress-importer.php
@@ -771,12 +771,16 @@ class WP_Import extends WP_Importer {
 				$num_comments = 0;
 				$inserted_comments = array();
 				foreach ( $post['comments'] as $comment ) {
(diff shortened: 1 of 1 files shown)


ocean90:
Thanks for submitting the PR!

The user ID is already set in https://github.com/WordPress/wordpress-importer/blob/8ff59108e81faff66d658e5ed592ba6d0364e252/src/wordpress-importer.php#L793-L794 based on the author mapping. Using users which occasionally match an existing user login is error-prone since it may not be the same user.

DesignerThan95:
I have used the Wordpress Unit Test data and wanted to use the authors defined in this file. That did not work. The comment had a user id = 0. Thats why I added those lines to the plugin.
But i will have a look on it later.

PR #75, opened 2020-04-14
Add PHPCS config and fix errors

szepeviktor:
Now we pass

```bash
vendor/bin/phpcs . --standard=WordPress-Core --exclude=WordPress.PHP.StrictComparisons,WordPress.WP.I18n
```

Restart of #31

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 dist: trusty
+
 language: php
 
 notifications:
@@ -8,11 +9,10 @@ notifications:
 
 cache:
   directories:
-    - vendor
     # Cache directory for older Composer versions.
-    - $HOME/.composer/cache/files
+    - $HOME/.composer/cache
     # Cache directory for more recent Composer versions.
-    - $HOME/.cache/composer/files
+    - $HOME/.cache/composer
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,11 @@ before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
   - phpenv config-rm xdebug.ini || echo 'No xdebug config.'
   - |
+    if [[ "$TRAVIS_PHP_VERSION" == "7.3" ]]; then
+      composer global require wp-coding-standards/wpcs phpcompatibility/phpcompatibility-wp dealerdirect/phpcodesniffer-composer-installer
+      phpcs
+    fi
+  - |
     bash phpunit/install.sh wordpress_test root '' localhost $WP_VERSION
     if [[ "$TRAVIS_PHP_VERSION" != "5.2" ]]; then
       composer global require "phpunit/phpunit=4.8.*|5.7.*|6.5.*|7.5.*"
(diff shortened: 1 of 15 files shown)


szepeviktor:
> WordPress.PHP.StrictComparisons

I do not dare to touch those as I don't know what actual values are compared.

> WordPress.WP.I18n

I'm not up to phrasing instructions for translators.

phpunit/tests/term-meta.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php // phpcs:ignore WordPress.Files.FileName
```
ocean90, line 1: This should be added as a exclude rule to the `phpunit.xml.dist` file.

src/class-wp-import.php
```diff
@@ -229,25 +231,25 @@ function get_authors_from_import( $import_data ) {
...
 		<li><?php $this->author_select( $j++, $author ); ?></li>
 <?php endforeach; ?>
 	</ol>
 <?php endif; ?>
 
-<?php if ( $this->allow_fetch_attachments() ) : ?>
+		<?php if ( $this->allow_fetch_attachments() ) : ?>
 	<h3><?php _e( 'Import Attachments', 'wordpress-importer' ); ?></h3>
 	<p>
 		<input type="checkbox" value="1" name="fetch_attachments" id="import-attachments" />
```
ocean90, line 255: This doesn't look correct, see https://github.com/szepeviktor/wordpress-importer/blob/phpcs/src/class-wp-import.php#L228-L263.

phpunit/tests/term-meta.php
szepeviktor: PHPCS does not provide a way to exclude rules for certain files only :(

What to do now?

src/class-wp-import.php
szepeviktor: Done.

phpunit/tests/term-meta.php
jrfnl: It most definitely does.

```xml
<rule ref="WordPress.Files.FileName">
    <exclude-pattern>*/phpunit/tests/*\.php$</exclude-pattern>
</rule>
```

Note: exclude patterns are treated like PCRE regexes with one caveat, that all `*`s are automatically converted to `.*`.

szepeviktor: Thank you @jrfnl!

Done.

.travis.yml
```diff
@@ -50,6 +50,11 @@ matrix:
 before_script:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
   - phpenv config-rm xdebug.ini || echo 'No xdebug config.'
+  - |
+    if [[ "$TRAVIS_PHP_VERSION" == "7.3" ]]; then
+      composer global require wp-coding-standards/wpcs phpcompatibility/phpcompatibility-wp dealerdirect/phpcodesniffer-composer-installer
+      phpcs
+    fi
```
ocean90, line 57: `before_script` shouldn't run any scripts.  It's better to have a separate job for phpcs.

szepeviktor: Done.

szepeviktor:
Thank you.

PR #76, opened 2020-05-15
Prevent duplicate postmeta records insertion

sagarsdeshmukh:
**Issue**
- On import of already existing attachment/post/page, the importer gives error "Page/Post/Attachment  already exists." but imports the postmeta records from source.

**Fix**
- Replace the function to add postmeta only if it's not exist

### DIFF
--- a/src/class-wp-import.php
+++ b/src/class-wp-import.php
@@ -895,7 +895,7 @@ class WP_Import extends WP_Importer {
 							$value = maybe_unserialize( $meta['value'] );
 						}
 
-						add_post_meta( $post_id, wp_slash( $key ), wp_slash_strings_only( $value ) );
+						update_post_meta( $post_id, wp_slash( $key ), wp_slash_strings_only( $value ) );
 
 						do_action( 'import_post_meta', $post_id, $key, $value );
 


dd32:
Hi @sagarsdeshmukh, I apologise for how long this PR has sat here without any comment.

Does swapping from `add_post_meta()` to `update_post_meta()` break the expectation that WordPress can have multiple post meta records per name? For example, you can have `name: animal value: cat` and `name: animal value: dog` attached to the same post.

It might be better to explicitly only import postmeta/taxonomies when the post itself is imported, rather than relying upon the behaviour of update vs add.

jmdodd:
Part of the problem here is that there can actually be multiple identical post meta records per name (maybe for counting thumbs-up/thumbs-down as an example use case):
```
wp> add_post_meta( 5, 'foo', 'bar' );
=> int(200)
wp> add_post_meta( 5, 'foo', 'bar' );
=> int(201)
wp> get_post_meta( 5, 'foo' );
=> array(2) {
  [0]=>
  string(3) "bar"
  [1]=>
  string(3) "bar"
}
```

If the import is run multiple times, using add_post_meta will result in multiple instances of each post meta on a post because these are imported regardless of whether the post previously existed, and independent of how many times they already exist on a post.

stevus:
@dd32 @sagarsdeshmukh Is there anything pending here? I'm attempting to import multiple WXR files produced by [WP_CLI export](https://developer.wordpress.org/cli/commands/export/) and running into this issue. Seems like a pretty straightforward fix? 

If I delete all postmeta for all pages / posts, and run an import using this tool on the same site that was used for the export, my site loads as expected afterwards. If I don't do a cleanup, all posts and pages have all of their postmeta duplicated.

stevus:
Should there be a check of existence of the key / value pair that is being imported, and if exists, do nothing? @sagarsdeshmukh @jmdodd that might solve the issue here?

sagarsdeshmukh:
This PR is being here for long period of time, I would like to ask the core team of this project, do you expect any action/input/changes from my end, please let me know? 
cc: @ocean90 @jffng @boonebgorges @dd32

PR #85, opened 2020-10-20
Update class-wp-import.php

BackuPs:
Please adapt the changes to import custom field created by plugins and themes,

Thank you


src/class-wp-import.php
```diff
@@ -953,7 +953,8 @@ function process_menu_item( $item ) {
 		} else {
 			$menu_id = is_array( $menu_id ) ? $menu_id['term_id'] : $menu_id;
 		}
-
+		$backup_menu_item_meta=array();
```
BackuPs, line 956: Save the current meta items in another temporary array

PR #87, opened 2020-12-15
Travis: add build against PHP 8.0

jrfnl:
PHP 8 was released a few weeks ago by now and Travis offers an `8.0` image, so adding a build against PHP 8.0 which is not allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
 
 matrix:
   include:
+    - php: 8.0
     - php: 7.4
       env: WP_VERSION=trunk
     - php: 7.4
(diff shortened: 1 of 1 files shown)


ocean90:
Since Travis no longer works in the WordPress organization I'm going to close this. We have to move to GitHub Actions.

jrfnl:
@ocean90 I understand, but it may not be a bad idea to merge it anyway, so whomever creates the GHA workflows is triggered to make sure that the tests run and pass on PHP 8.0.

ocean90:
Created a new issue #91 so it doesn't get missed.

PR #92, opened 2020-12-15
Tests: improve test quality

jrfnl:
:point_right: This PR will be easiest to review by looking at each commit individually.

### PHPUnit: tweak configuration

* Add XML annotation.
* Add schema annotation.
* Add code coverage configuration.

### Tests: fix incorrect parameter order

All PHPUnit assertions which take both an `$expected` as well as an `$actual` value, have the `$expected` parameter as the first and the `$actual` as the second parameter.

While passing the parameters in reverse order will work, it will make the error message when the assertion fails confusing.

This fixes the few instances where the order was reversed.

### Tests: use `assertCount()`

Use the most appropriate assertion available in PHPUnit instead of working things out in the tests themselves.

### Tests: use `assertInternalType()`

Use the most appropriate assertion available in PHPUnit instead of working things out in the tests themselves.

Note: `assertInternalType()` is deprecated in PHPUnit 7 in favour of more specific assertions, like `assertIsArray()`, but as these tests, for now, won't be run on PHPUnit > 7, this is not an issue.

### Tests: use the most specific assertion

PHPUnit contains a wealth of assertions, even in PHPUnit 3.x, which this plugin still needs to use, so use the most specific assertion possible.

### Tests: use `assertSame()` instead of `assertEquals()` [1]

Improve test quality.

This changes calls to `assertEquals()` to `assertSame()` when the `$expected` value is a text string.

`assertEquals()` does a loose type comparison and `assertSame()` does a strict type comparison.

Using strict type comparisons should always be preferred.
This is especially important for assertions against strings as strings would juggle to any other type, so a loose type comparison may incorrectly pass.

Refs:
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertEquals
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertSame

### Tests: use `assertSame()` instead of `assertEquals()` [2]

Improve test quality.

This changes calls to `assertEquals()` to `assertSame()` when the `$expected` value is a non-text string or a variable.

`assertEquals()` does a loose type comparison and `assertSame()` does a strict type comparison.

Using strict type comparisons should always be preferred.

Refs:
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertEquals
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertSame

### Tests: use `assertSame()` instead of `assertEquals()` [3]

Improve test quality.

This changes calls to `assertEquals()` to `assertSame()`. In all these cases, the _type_ of the value being tested was incorrect.

Changing the code to return the correct type should be considered a BC-break. For this reason, I've elected to update the tests instead, documenting what type of value to expect.

---

### 🆕 Tests: revert a `assertSame()` change and annotate why

Based on the test results, the property type of `$comments_count->approved` has changed across WP versions from string to integer.


jrfnl:
Now we have CI working again, it becomes clear that the stricter tests would fail on a select number of PHP - WP combinations.

Aside from one change which I've reverted, the remaining four build failures (same three test failures) are all low PHP - WP combinations, which would be dropped if/when #95 would get actioned.

With that in mind, I propose getting #95 sorted first and rebasing and merging this PR afterwards.

jrfnl:
@ocean90 Would you mind giving feedback please ? Adding commits to someone else's PR without communicating about it makes things murky.

ocean90:
I think the commit is clear enough.

jrfnl:
@ocean90 That's not the point.

jrfnl:
PR #99 has been opened to address issue #95. Once that's merged, I'll rebase this PR (and expect it to pass).

jrfnl:
Rebased and the build is now 🟢

ocean90:
Nice work, thanks!

PR #93, opened 2020-12-28
CI: switch to GitHub Actions

### LINKED ISSUE #91: Replace Travis CI with GitHub Actions
Since Travis CI no longer runs builds for open source projects, the current [`.travis.yml`](https://github.com/WordPress/wordpress-importer/blob/7bbdcdfddd1ceecc6ec3fa2735aec3f11f7a88cb/.travis.yml#L1) config needs to be converted to one or more GitHub workflows.

With the switch we should also make sure that all current PHP versions are tested.

jrfnl:
This switches out the Travis script for GitHub actions.

~~**Known issue**:~~
~~It appears that the MySQL version being set in the `services -> image` part is not being respected as can be verified via the `Verify mysql version` step.~~
~~No matter what I do, MySQL 5.7 will be installed. This is, at least in part, the reason some of the test runs against WP 3.7 are currently failing.~~

~~Any ideas on how to fix this would be appreciated!~~

Fixes #91

---

### CI: switch to GitHub Actions - step 1: sniff stage

This commit:
* Adds a GH Actions workflow for the code style related CI checks.
* Removes that part of the `.travis.yml` configuration.
* Adds a "Build Status" badge in the Readme to use the results from this particular GH Actions runs.

Notes:
Builds will run on all pushes and on pull requests, with the exception of those just modifying files which are irrelevant to this workflow.

### Raise minimum required PHP version to PHP 5.3

As:
* Testing on PHP 5.2 is becoming harder and harder (especially as the Setup PHP action does not support PHP 5.2);
* The PHP project itself not having supported PHP 5.2 since January 2011;
* And the WP project having dropped support for PHP < 5.6 in WP 5.2, which was released in May 2019...

I propose to drop support for PHP 5.2 and make the new minimum supported PHP version PHP 5.3. This is still exceedingly liberal, but at least is a version for which support can be verified by running the tests.

Includes tidying up the file docblock in the `wordpress-importer.php` file.

### PHPUnit: make WP install script executable

Change the file permissions for the `phpunit/install.sh` to allow execution of the script.

### CI: switch to GitHub Actions - step 2: tests

This commit:
* Adds a GH Actions workflow for the test runs which were previously run on Travis.
* Removes the, now redundant, `.travis.yml` configuration.
* Updates the `.gitattributes` file.
* Updates the "Build Status" badge in the Readme to show the results from the GH `Test` Actions runs.

Notes:
1. In the build matrix on Travis, PHP 7.0 - 7.4 were tested against the "latest" WP version and a few select WP versions were tested against lower PHP versions.
    In the proposed workflow for GH Actions, the matrix is made much more complete with:
    - PHP 5.6 up to 8.0 being tested against both the lowest (`3.7`) and highest (`latest`) officially supported WP version for this plugin.
    - PHP 5.3 up to 5.5 being tested against the lowest (`3.7`) and the highest (`5.1`) WP version which will allow still run on these PHP versions.
    - Selectively testing against some interim WP versions.
2. For the MySQL version, we need to take into account when support for which MySQL version was added to WP.
    - WP 4.5 is the first WP version which properly supports MySQL 5.7.
    - WP 5.4 is the first WP version which largely supports MySQL 8.0.
3. It appears that the PHP 5.6 and 7.0 images do not natively come with the `mysql` or `mysqli` extensions.
    For that reason, the extension is explicitly added.
    Refs:
    * https://github.com/shivammathur/setup-php/wiki/Php-extensions-loaded-on-ubuntu-18.04#php-56
    * https://github.com/shivammathur/setup-php/wiki/Php-extensions-loaded-on-ubuntu-18.04#php-70
4. As these tests are integration tests with the WP native test suite, we need to take the PHPUnit version supported by WP into account for all supported WP versions.
    - Originally only PHPUnit 4.x and 5.x were supported.
    - As of WP 5.1, PHPUnit 6.x and 7.x are also supported.
    - With WP 5.2 dropping support for PHP < 5.6, support for PHPUnit 4.x was also dropped.
    - On PHP >= 8.0, PHPUnit 7.5+ installed via Composer using `--ignore-platform-reqs` is needed, no matter what.

---

### GH Actions: allow for manually triggering a workflow

Triggering a workflow for a branch manually is not supported by default in GH Actions, but has to be explicitly allowed.

Ref: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### GH Actions: report CS violations in the PR

The cs2pr tool will allow to display the results from an action run in checkstyle format in-line in the PR code view, which should improve usability of the workflow results.

Ref: https://github.com/staabm/annotate-pull-request-from-checkstyle

### GH Actions: fix running against various MySQL versions

These seemingly minor tweaks allow for installing different MySQL versions in the local VM depending on the variables set in the matrix.

This fixes the previously seen test failures when the tests were being run against MySQL 5.7 with an incompatible WP version.

### GH Actions: limit the use of MySQL 8.0 to PHP 7.4 or higher

... as based on the test output, there are issues in WP Core which prevent the tests from being run against MySQL 8.0 using PHP 7.3 or lower.

These issues have been reported upstream. See: https://core.trac.wordpress.org/ticket/52496

### GH Actions: don't test against PHP nightly

WP Core limits the PHPUnit version used for the test suite to PHPUnit 5.7 - 7.5, neither or which are still supported by PHPUnit.

These PHPUnit versions contain an incompatibility with PHP 8.1 which will not be fixed anymore (as the PHPUnit versions are no longer supported).

Until a solution for this has been found and applied to the WP Core test suite - be it widening the supported PHPUnit version range or otherwise -, running the tests against PHP 8.1/nightly is useless as they will fail on the PHPUnit incompatibility without ever running the plugin tests.


.github/workflows/cs.yml
```diff
@@ -0,0 +1,41 @@
+name: CS
```
desrosj, line 1:
```suggestion
name: Coding Standards
```

desrosj: Is there any reason to not be more explicit here? Since this is displayed in the UI, it may help to write out the full name.

.github/workflows/test.yml
```diff
@@ -0,0 +1,183 @@
...
+          # Experimental builds. These are allowed to fail.
+          - php: '7.4'
+            wp: 'trunk'
+            experimental: true
+
+          - php: '8.1'
+            wp: 'trunk'
+            experimental: true
+
+    name: "Test: PHP ${{ matrix.php }} - WP ${{ matrix.wp }}"
```
desrosj, line 100: Don't feel strongly about this, but it always feels repetitive when naming jobs to include the name of the workflow in the job name. Test is implied here from the workflow name.

.github/workflows/test.yml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Install Composer dependencies for PHP >= 8.0
+        if: ${{ matrix.php >= 8.0 }}
+        uses: "ramsey/composer-install@v1"
+        with:
+          composer-options: --ignore-platform-reqs
+
+      - name: Run the unit tests - single site
+        run: vendor/bin/phpunit
+
+      - name: Run the unit tests - multi site
```
desrosj, line 180:
```suggestion
      - name: Run the unit tests - multisite
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,183 @@
...
+          mysqld --version
+          mysqladmin -V
+
+      - name: Set up WordPress
+        run: phpunit/install.sh wordpress_test root root 127.0.0.1 ${{ matrix.wp }}
+
+      - name: Determine supported PHPUnit version
+        id: set_phpunit
+        run: |
+          # Originally, only PHPUnit 4.x and 5.x were supported.
```
desrosj, line 145: If the comment is included as a part of the `run` command, it will be printed out to the [first line of the step](https://github.com/WordPress/wordpress-importer/runs/1615689255?check_suite_focus=true#step:8:1).

Not a huge deal, but this docblock could be moved above the name to make the step's output a bit cleaner.

desrosj (commented):
Overall this looks good! Left a few small comments.

I also tried experimenting a bit to fix the MySQL version issue but haven't figured it out just yet. My hunch is that it has something to do with either the ports, or how `mysql` is being started/named. I don't _think_ that we need to necessarily start the service because the [container should be running after it's checked out](https://github.com/WordPress/wordpress-importer/runs/1615689328?check_suite_focus=true#step:2:95). I'll keep searching and testing, but may be worth posting in the [GitHub Actions community](https://github.community/c/code-to-cloud/github-actions/41) for clarification.

desrosj:
Also, as for the PHP version change, I'm not sure the process we'd need to follow to make the supported version change officially.

While I appreciate the problems with testing PHP 5.2 (currently battling to get the Core tests to run on 5.2 myself), I'm not sure that we can drop support for 5.2 without the potential to break older sites. We could _maybe_ make that change if we also raised the minimum WordPress version. I believe that this would prevent sites from updating to a new version of the plugin potentially containing breaking changes on older versions of PHP. Another option may be to include some form of PHPCompatability checking that scans for PHP 5.2 issues as a way to verify compatibility instead of unit testing, though that is of course no where near the same thing.

jrfnl:
@desrosj Thanks for looking this over. FYI: I will get back to this and update the PR in a few weeks (once I have my new PC up and running).

.github/workflows/cs.yml
jrfnl: The reason - for me - was that the name in the button on the Readme then becomes very very long, but I'm not fussed either way.

.github/workflows/test.yml
jrfnl: Fair enough. As this plugin doesn't have separate "quick test" or code coverage runs which could be using a similar name pattern, having the added `Test` is unnecessary.

I would note though that as soon as either of these are added, a keyword to distinguish between the various test run times would be needed, but that's for later (if at all).

jrfnl:
> Also, as for the PHP version change, I'm not sure the process we'd need to follow to make the supported version change officially.
>
> While I appreciate the problems with testing PHP 5.2 (currently battling to get the Core tests to run on 5.2 myself), I'm not sure that we can drop support for 5.2 without the potential to break older sites. We could maybe make that change if we also raised the minimum WordPress version. I believe that this would prevent sites from updating to a new version of the plugin potentially containing breaking changes on older versions of PHP. Another option may be to include some form of PHPCompatability checking that scans for PHP 5.2 issues as a way to verify compatibility instead of unit testing, though that is of course no where near the same thing.

Some notes about this:

The plugin no longer being tested on PHP 5.2 doesn't mean that it stops working on PHP 5.2. It just means that extra care should be taken when reviewing commits to make sure no code enters the code base which is incompatible with PHP 5.2.

As for officially dropping support for PHP 5.2:
This commit does add a `Requires PHP`. The header as been supported for quite a while: https://make.wordpress.org/plugins/2017/08/29/minimum-php-version-requirement/ (August 2017), though enforcement when installing/updating plugins looks to have only been introduced - partially - in WP 5.1 (?) and partially in WP 5.5.
Refs:
- https://core.trac.wordpress.org/ticket/40934
- https://core.trac.wordpress.org/changeset/43436
- https://core.trac.wordpress.org/ticket/40934#comment:105

So basically that's not enough.

That leaves us with a choice between the following:
1. Drop support for WP < 5.2, which has a minimum PHP requirement of PHP 5.6.
2. Add a PHP version check to the plugin which will auto-deactivate the plugin when the minimum PHP requirement of PHP 5.3 has not been met and adds an admin notice informing the user.
3. Figure out some way to still get the tests running on PHP 5.2.

I think this needs to be discussed and addressed in a separate ticket though.

This ticket addresses the fact that **_no QA checks at all_** are currently being run. Running the QA checks for every single PHP version, but 5.2 is still _much_ better than not having any QA checks at all.

The fact that QA is not executed on PHP 5.2 and what to do about that, can be discussed after this PR has been merged (and before a next release).

@desrosj Do you agree with this approach ? Shall I open a new ticket for this ?

dd32:
> * Drop support for WP < 5.2, which has a minimum PHP requirement of PHP 5.6.
> * Add a PHP version check to the plugin which will auto-deactivate the plugin when the minimum PHP requirement of PHP 5.3 has not been met and adds an admin notice informing the user.

This would be my suggestion - do both of these, drop support for older WP/PHP versions, if anyone on older versions needs old version support, they can install an older version of the plugin.

jrfnl:
@dd32 Thanks for your input. I'd suggest we do so in a separate PR, though dropping support for WP < 5.2 will simplify the matrix used here _a lot_.

jrfnl:
@desrosj Just checking - will this repo have "required statuses" setup for PRs ?
If so, I should probably remove the `paths-ignore` bits preventing a run on `txt`/`md`-only changes, as for documentation only PRs the workflows not being run would block those PRs from being merged in that case.

jrfnl:
Right, with some more trial & error, I now have the testing against different MySQL versions working. I will clean up the (trial and error) commits in the test branch and add the result here in a moment.

While doing this, I did find that - [based on the test output](https://github.com/WordPress/wordpress-importer/actions/runs/556887133) - WP Core apparently is not properly compatible with MySQL 8.0, except when used in combination with PHP 7.4+.

I do a write up of the findings and open a Trac ticket about this in WP Core.

jrfnl:
Pr description has been updated with the information on the additional commits & we now have a passing build 🟢

jrfnl:
I've opened #95 to discuss and document the dropping of support for WP < 5.2 / PHP < 5.6.

jrfnl:
Workflow wise - I'd like to propose getting this PR + PR #92 merged first before taking action on #95, so the drop can be done cleanly without having to redo already open PRs.

Issue #41, opened 2018-08-20
gzopen(): cannot represent a stream of type user-space as a File Descriptor


alexjurkiewicz:
Using version 0.6.4 with Wordpress 4.9.8.

Received this error when trying to import any content on our Wordpress:

```
[Mon Aug 20 02:17:12.570185 2018] [php7:warn] [pid 18589] [client 10.120.64.239:57886] PHP Warning:  gzopen(): cannot represent a stream of type user-space as a File Descriptor in /var/app/current/wp-content/plugins/wordpress-importer/parsers.php on line 677, referer: https://our.site.com/wp-admin/admin.php?import=wordpress
```

Commenting out lines [676 & 677](https://github.com/WordPress/wordpress-importer/blob/9a1aed1ccbf2b513cf0387b069f163e78bd937e7/src/parsers.php?utf8=%E2%9C%93#L676) allows us to import (uncompressed) dumps just fine.

PR #96, opened 2021-02-11
Using gzip is optional.

fumikito:
In some environments like a shared folder, gzip is not reliable even if PHP has Gzip extension.
Related #41


jrfnl:
@fumikito Looks like a change which would be perfectly acceptable to add. Would you mind adding a few tests covering the various scenarios ?

fumikito:
@jrfnl Okay, I'll check the tests and add extra commits to this PR! Wait for a while.

PR #97, opened 2021-02-13
Preserve blank lines in posts during import

dsnyder0pc:
Due to a logic error, blank lines were unintentionally skipped during imports. This change explicitly skips tag lines while preserving blank lines in posts.

PR #98, opened 2021-02-15
Regex parser: preserve blank lines in posts

### LINKED ISSUE #94: Blank lines skipped in posts
I've observed that when importing posts with (intentional) embedded blank lines, the posts show up in the editor with the blank lines removed. I'm wondering if [this code change](https://github.com/WordPress/wordpress-importer/compare/0.6.4...0.7#diff-fbded9ce0f4cf9659a6895ef97e4e321ea4f4baaca594c6e12b66c4964dbcbbcR87-R89) is what is causing that:

import xml file

Here's how this looks in the WordPress editor for "Free" customers on WordPress.com:

import xml file

As you can see, the blank lines are preserved. However, when I import the same content to a "Business" plan on WordPress.com, the blank lines appear to have been filtered out:

import xml file

My hypothesis is that, for blank lines, the PHP language evaluates $importline to False, causing the importer to skip over line 88. I have not tested this but may do so later this week. If so, I'll report back.

I'm in the process of importing 2,100 posts into a Business plan on WordPress.com, and the posts are basically unreadable with the missing whitespace. Thanks for having a look.

yoavf:
This is a follow up on #97, which adds a test case to make sure all the parsers preserve blank lines in content
The test demonstrates the issue reported in #94, and the change by @dsnyder0pc fixes it.

Fixes #94


ocean90:
Thanks for the test! I think it's better to change the base branch to #97 so we can merge both changes as one.

PR #99, opened 2021-02-15
Raise minimum required WP version to WP 5.2 ( and PHP 5.6)

### LINKED ISSUE #95: Drop support for WP < 5.2 / PHP < 5.6
Continuing to support the current huge range of supported PHP and WP combinations is starting to become more and more difficult, especially from a CI/testing perspective.

With that in mind, I am proposing we drop support for PHP < 5.6 and WP < 5.2 in the next "major" release (`0.8`).
Users on older WP/PHP versions can continue to use previous (older) releases of the plugin.

### Relevant parts of a discussion about this in #93:

---

@desrosj :
> While I appreciate the problems with testing PHP 5.2 (...), I'm not sure that we can drop support for 5.2 without the potential to break older sites. We could maybe make that change if we also raised the minimum WordPress version. I believe that this would prevent sites from updating to a new version of the plugin potentially containing breaking changes on older versions of PHP. Another option may be to include some form of PHPCompatability checking that scans for PHP 5.2 issues as a way to verify compatibility instead of unit testing, though that is of course no where near the same thing.

---
 @jrfnl in response to this:

> Some notes about this:
> 
> The plugin no longer being tested on PHP 5.2 doesn't mean that it stops working on PHP 5.2. It just means that extra care should be taken when reviewing commits to make sure no code enters the code base which is incompatible with PHP 5.2.
> 
> As for officially dropping support for PHP 5.2:
> This commit does add a `Requires PHP`. The header has been supported for quite a while: https://make.wordpress.org/plugins/2017/08/29/minimum-php-version-requirement/ (August 2017), though enforcement when installing/updating plugins looks to have only been introduced - partially - in WP 5.1 (?) and partially in WP 5.5.
> Refs:
> - https://core.trac.wordpress.org/ticket/40934
> - https://core.trac.wordpress.org/changeset/43436
> - https://core.trac.wordpress.org/ticket/40934#comment:105
> 
> So basically that's not enough.
> 
> That leaves us with a choice between the following:
> 1. Drop support for WP < 5.2, which has a minimum PHP requirement of PHP 5.6.
> 2. Add a PHP version check to the plugin which will auto-deactivate the plugin when the minimum PHP requirement of PHP 5.3 has not been met and adds an admin notice informing the user.
> 3. Figure out some way to still get the tests running on PHP 5.2.
> 
> The fact that QA is not executed on PHP 5.2 and what to do about that, can be discussed after this PR has been merged (and before a next release).

----

@dd32 in response:
> > * Drop support for WP < 5.2, which has a minimum PHP requirement of PHP 5.6.
> > * Add a PHP version check to the plugin which will auto-deactivate the plugin when the minimum PHP requirement of PHP 5.3 has not been met and adds an admin notice informing the user.
> 
> This would be my suggestion - do both of these, drop support for older WP/PHP versions, if anyone on older versions needs old version support, they can install an older version of the plugin.

jrfnl:
### Raise minimum required WP version to WP 5.2 ( and PHP 5.6)

Includes:
* Adding `Requires PHP` headers.
* Updating the `Requires at least` (WP) headers.
* Stop detection of PHP compatibility issues for PHP < 5.6.
* CI: Stop testing on WP < 5.2 and PHP < 5.6.
    Note: the PHP 7.4 and 8.0 builds against WP 5.2 were running into trouble due to WP itself not being fully compatible, so the CI matrix has been adjusted to only test against those PHP versions on WP versions which claim to be compatible with those PHP versions.

Includes tidying up the file docblock in the `wordpress-importer.php` file.

Closes #95

### Clean up: remove work-arounds and polyfills

With WP 5.2 being the new minimum supported WP version, the polyfill for the `map_deep()` function is no longer needed.

Similarly, the bowing out for the term meta import (on WP < 4.4) is also no longer needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,72 +22,34 @@ jobs:
       fail-fast: false
       matrix:
         # Notes regarding supported versions in WP:
-        # PHP 5.3-5.5 are supported since WP 3.7 until WP 5.1 (including).
-        # PHP 5.6 is supported since WP 4.1.
-        # PHP 7.0 is supported since WP 4.4.
-        # PHP 7.1 is supported since WP 4.7.
-        # PHP 7.2 is supported since WP 4.9.
-        # PHP 7.3 is supported since WP 5.0.
-        # PHP 7.4 is supported since WP 5.3.
-        # PHP 8.0 is sort of supported since WP 5.6.
-        # This doesn't mean that the plugin doesn't work on earlier version as it may not hit incompatible
-        # parts of WP, but just something to be aware of in case of failures.
-        php: ['5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0']
-        wp: ['latest', '3.7']
+        # The base matrix only contains the PHP versions which are supported on all supported WP versions.
+        php: ['5.6', '7.0', '7.1', '7.2', '7.3']
+        wp: ['latest', '5.2']
         experimental: [false]
 
         include:
-          # As support for PHP < 5.6 was dropped in WP 5.2, PHP 5.3 - 5.5 can only be tested against WP < 5.2.
-          # Complement the builds run via the matrix with high/low WP builds for PHP 5.3 - 5.5.
-          - php: '5.5'
-            wp: '5.1'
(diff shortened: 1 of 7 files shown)


ocean90 (approved):
Let's do this.

PR #100, opened 2021-03-10
Update class-wp-import.php

rvpatel:
https://core.trac.wordpress.org/ticket/52740
#52740


lkraav:
I'm fairly certain this should be a CSS solution.

PR #105, opened 2021-08-19
Fix conditional logic for setting base_blog_url

jjpmann:
don't override base_blog_url if already set

### DIFF
--- a/src/parsers/class-wxr-parser-regex.php
+++ b/src/parsers/class-wxr-parser-regex.php
@@ -55,7 +55,7 @@ class WXR_Parser_Regex {
 					preg_match( '|<wp:base_blog_url>(.*?)</wp:base_blog_url>|is', $importline, $blog_url );
 					$this->base_blog_url = $blog_url[1];
 					continue;
-				} else {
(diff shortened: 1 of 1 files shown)


dd32:
Hi @jjpmann,
Thanks for the PR!

Looking at the change here, it seems reasonable, can you expand upon the need for it / if there's a ticket explaining the bug somewhere?

Thanks!

jjpmann:
Hi @dd32 ,

Sorry for missing the details. I created a ticket here but didn't exactly know the full process. 
https://core.trac.wordpress.org/ticket/53956

Basically, if the regex parse was used the blog_url would always get overridden unless it was the last parsed line in the loop.

dd32:
Reviewing this closer..

SimpleXML has similar logic, but runs once prior to the parse, rather than each loop.
https://github.com/WordPress/wordpress-importer/blob/master/src/parsers/class-wxr-parser-simplexml.php#L55-L63

XML Parser has similar logic, only setting once thanks to `isset()`:
https://github.com/WordPress/wordpress-importer/blob/master/src/parsers/class-wxr-parser-xml.php#L240-L244

Change seems appropriate.

PR #113, opened 2021-10-25
PHP 8.1: fix a "passing null to non-nullable" deprecation

jrfnl:
Fixes PHP 8.1 `basename(): Passing null to parameter #1 ($path) of type string is deprecated` error due to the call to `parse_url()` returning `null` when the component requested does not exist in the passed URL, which in this case means a domain without trailing slash.

Includes a new unit test which would error out without this fix.

Note: the new test does not actually test the functioning of the `WP_Import::fetch_remote_file()` method in any significant way. It is only a regression test to prevent this particular PHP 8.1 deprecation notice from being reintroduced.

### DIFF
--- a/phpunit/tests/import.php
+++ b/phpunit/tests/import.php
@@ -275,5 +275,22 @@ class Tests_Import_Import extends WP_Import_UnitTestCase {
 		$this->assertSame( '\o/ ¯\_(ツ)_/¯', $comments[0]->comment_content );
 	}
 
+	/**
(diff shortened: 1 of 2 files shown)


jrfnl:
As WP 5.8.2 has been released, I've re-run the test workflow for this PR and all is now 🟢 (save for PHP 8.1 - [see my explanation above](https://github.com/WordPress/wordpress-importer/pull/113#issuecomment-951145180), but that issue is not related to this PR). As far as I'm concerned, let's get this merged.

PR #121, opened 2022-01-30
Pass true as the second parameter to wp_insert_category()

mauriac:
Trac ticket: https://core.trac.wordpress.org/ticket/54973


jrfnl:
Looks like this was fixed via #123

PR #128, opened 2022-06-17
PHP 8.2 | Explicitly declare all properties

jrfnl:
Dynamic (non-explicitly declared) property usage is expected to be deprecated as of PHP 8.2 and are expected to become a fatal error in PHP 9.0.

There are a number of ways to mitigate this:
* If it's an accidental typo for a declared property: fix the typo.
* For known properties: declare them on the class.
* For unknown properties: add the magic `__get()`, `__set()` et al methods to the class or let the class extend `stdClass` which has highly optimized versions of these magic methods build in.
* For unknown _use of_ dynamic properties, the `#[AllowDynamicProperties]` attribute can be added to the class. The attribute will automatically be inherited by child classes.

Refs:
* https://wiki.php.net/rfc/deprecate_dynamic_properties

### WXR_Parser_XML: explicitly declare all properties [1]

In this case, the `parse()` method explicitly sets each of those properties, so these fall in the "known property" category and should be explicitly declared.

### WXR_Parser_XML: explicitly declare all properties [2]

In this case, the `tag_close()` method explicitly sets each of those properties, so these fall in the "known property" category and should be explicitly declared.

### WXR_Parser_Regex: explicitly declare all properties

In this case, the `__construct()` method explicitly sets this property, so `$has_gzip` falls in the "known property" category and should be explicitly declared.

### DIFF
--- a/src/parsers/class-wxr-parser-regex.php
+++ b/src/parsers/class-wxr-parser-regex.php
@@ -10,13 +10,14 @@
  * WXR Parser that uses regular expressions. Fallback for installs without an XML parser.
  */
 class WXR_Parser_Regex {
-	var $authors       = array();
-	var $posts         = array();
-	var $categories    = array();
-	var $tags          = array();
-	var $terms         = array();
-	var $base_url      = '';
-	var $base_blog_url = '';
+	public $authors       = array();
+	public $posts         = array();
+	public $categories    = array();
+	public $tags          = array();
+	public $terms         = array();
+	public $base_url      = '';
+	public $base_blog_url = '';
+	public $has_gzip;
 
 	function __construct() {
 		$this->has_gzip = is_callable( 'gzopen' );
--- a/src/parsers/class-wxr-parser-xml.php
+++ b/src/parsers/class-wxr-parser-xml.php
@@ -70,6 +70,8 @@ class WXR_Parser_XML {
 	public $term;
 	public $category;
 	public $tag;
+	public $base_url;
+	public $base_blog_url;
 
 	function parse( $file ) {
 		$this->wxr_version = false;
--- a/src/parsers/class-wxr-parser-xml.php
+++ b/src/parsers/class-wxr-parser-xml.php
@@ -10,7 +10,7 @@
  * WXR Parser that makes use of the XML Parser PHP extension.
  */
 class WXR_Parser_XML {
-	var $wp_tags     = array(
+	public $wp_tags     = array(
 		'wp:post_id',
 		'wp:post_date',
 		'wp:post_date_gmt',
@@ -43,7 +43,7 @@ class WXR_Parser_XML {
 		'wp:author_first_name',
 		'wp:author_last_name',
 	);
-	var $wp_sub_tags = array(
+	public $wp_sub_tags = array(
 		'wp:comment_id',
 		'wp:comment_author',
 		'wp:comment_author_email',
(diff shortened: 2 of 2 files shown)


jrfnl:
P.S.: not claiming completeness with this PR, but these were the obvious ones I found straight off. These may be more to find though.

jrfnl:
N.B.: The build failure against PHP 8.1 is unrelated to this PR.

dd32 (approved):
As an initial patch for this, this seems safe to merge.

I do ponder if we should explicitly set default values for readability, even though the code would ignore that though? for example, setting `has_gzip` to `false` by default.

jrfnl:
> As an initial patch for this, this seems safe to merge.
> 
> I do ponder if we should explicitly set default values for readability, even though the code would ignore that though? for example, setting `has_gzip` to `false` by default.

I wondered the exact same thing, but as the value would be overwritten anyway, I figured it wasn't worth spending the time to determine what the defaults _should_ be.

src/parsers/class-wxr-parser-regex.php
```diff
@@ -10,13 +10,14 @@
...
-	var $base_url      = '';
-	var $base_blog_url = '';
+	public $authors       = array();
+	public $posts         = array();
+	public $categories    = array();
+	public $tags          = array();
+	public $terms         = array();
+	public $base_url      = '';
+	public $base_blog_url = '';
+	public $has_gzip;
```
dd32, line 20:
```suggestion
	public $has_gzip = false;
```

src/parsers/class-wxr-parser-xml.php
```diff
@@ -58,6 +58,21 @@ class WXR_Parser_XML {
...
+	public $sub_data;
+	public $in_tag;
+	public $in_sub_tag;
+	public $authors;
+	public $posts;
+	public $term;
+	public $category;
+	public $tag;
+	public $base_url;
+	public $base_blog_url;
```
dd32, line 74:
```suggestion
	public $wxr_version   = false;
	public $in_post       = false;
	public $cdata         = false;
	public $data          = false;
	public $sub_data      = false;
	public $in_tag        = false;
	public $in_sub_tag    = false;
	public $authors       = array();
	public $posts         = array();
	public $term          = array();
	public $category      = array();
	public $tag           = array();
	public $base_url      = '';
	public $base_blog_url = '';
```

dd32 (commented):
These seem like a good default to me, some things like `sub_data` are `array()` but there's specific handling for `false` too..

I don't think adding these makes a huge difference, as they'll all be overridden.. but I'll leave it up to you :)

jrfnl:
> I don't think adding these makes a huge difference, as they'll all be overridden.. but I'll leave it up to you :)

Well, the current patch does not break BC, while adding defaults potentially _could_ break BC, as a check with `isset()` for one of these properties will no longer return `false`.

As the test suite for this plugin is minimal, I would prefer to err on the side of caution and not run the risk of breaking BC.

jrfnl:
Note; not breaking BC is also the reason why I made these properties all `public`, even though, purely on their (known) use/merit, each of these properties should be `private` or at most `protected`.

dd32:
> Well, the current patch does not break BC, while adding defaults potentially could break BC

While I doubt there would be any BC issues, it ultimately makes almost no difference here to skip documenting these. Merged as-is.

PR #130, opened 2022-07-13
GH Actions: add build against PHP 8.2

jrfnl:
Now PR #128 has been merged, let's enable a GH Actions build against PHP 8.2.

While the build - for now - won't pass due to a PHP 8.1 issue with WP Core, I have done a test run which [didn't convert deprecations to exceptions](https://github.com/WordPress/wordpress-importer/runs/7320373532?check_suite_focus=true), and that run shows that the tests for the plugin itself will now pass.

Note: there are still plenty of deprecations coming from WP Core itself. Those are outside the scope of this ticket and not solvable here anyway.
The following WP Core tickets are open to address those:
* https://core.trac.wordpress.org/ticket/56033
* https://core.trac.wordpress.org/ticket/56034

~~Also note there is currently an issue with the GD extension on PHP 8.2. This - again - is outside the scope of this ticket. This is most likely an issue with one of the underlying libraries used by the `setup-php` action runner and is expected to be a temporary hickup.~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,9 @@ jobs:
           - php: '8.1'
             wp: 'trunk'
             experimental: true
+          - php: '8.2'
+            wp: 'trunk'
+            experimental: true
           - php: '7.4'
             wp: 'trunk'
             experimental: true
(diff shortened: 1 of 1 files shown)


jrfnl:
> Also note there is currently an issue with the GD extension on PHP 8.2. This - again - is outside the scope of this ticket. This is most likely an issue with one of the underlying libraries used by the setup-php action runner and is expected to be a temporary hickup.

The PHP 8.2 / GD extension hickup has been resolved and I have rerun the build now. The other notes still stand.

PR #131, opened 2022-07-14
Tests/GH Actions: allow for upstream issue which prevented the PHP 8.1 builds from passing

jrfnl:
### Tests: temporarily skip a test until WP updates the Requests library

This should allow the PHP 8.1 build to pass.

### GH Actions: do not allow the PHP 8.1 build to fail anymore

... and test against the highest and lowest support WP version.

### GH Actions: update the matrix for latest WP/PHP

### DIFF
--- a/phpunit/tests/import.php
+++ b/phpunit/tests/import.php
@@ -281,6 +281,11 @@ class Tests_Import_Import extends WP_Import_UnitTestCase {
 	 * Note: this test doesn't test anything else of the functionality in the `WP_Import::fetch_remote_file()` method!
 	 */
 	public function test_fetch_remote_file_php81_deprecation() {
+		// Temporary until WP updates to Requests 2.0.0.
+		if ( PHP_VERSION_ID >= 80100 ) {
+			$this->markTestSkipped( 'Test will fail on PHP 8.1+ until WP has upgraded to Requests 2.0.0. Temporarily skipping the test.' );
+		}
+
 		$importer = new WP_Import();
 		$result   = $importer->fetch_remote_file( 'https://example.com', array() );
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,10 +24,17 @@ jobs:
         experimental: [false]
 
         include:
-          # Complement the builds run via the matrix with high/low WP builds for PHP 7.4 and 8.0.
-          # PHP 8.1 is sort of supported since WP 5.9.
+          # Complement the builds run via the matrix with high/low WP builds for PHP 7.4 up to 8.2.
+          # PHP 8.2 is sort of supported since WP 6.1.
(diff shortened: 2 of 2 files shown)


jrfnl:
Once WP Core ticket https://core.trac.wordpress.org/ticket/54504 has been actioned, the temporary test skipping for the one test can be removed again.

As for the PHP 8.2 tests showing deprecation notices - all of those are related to upstream ticket https://core.trac.wordpress.org/ticket/56034

phpunit/tests/import.php
```diff
@@ -281,6 +281,11 @@ public function test_slashes_should_not_be_stripped() {
 	 * Note: this test doesn't test anything else of the functionality in the `WP_Import::fetch_remote_file()` method!
 	 */
 	public function test_fetch_remote_file_php81_deprecation() {
+		// Temporary until WP updates to Requests 2.0.0.
+		if ( PHP_VERSION_ID >= 80100 ) {
+			$this->markTestSkipped( 'Test will fail on PHP 8.1+ until WP has upgraded to Requests 2.0.0. Temporarily skipping the test.' );
```
dd32, line 286: This doesn't seem like the best fix - Yes, it'll make the tests pass, but that's it.

We can wait on trunk to fix the notices, or add a workaround into the test structure that allows for discarding the deprecation notices.

dd32 (changes requested):

phpunit/tests/import.php
jrfnl: Well, the fix for Core (upgrade Requests) has been available since November 2021 and it doesn't look like anything is happening there for now, so that can be a long wait.

The reason I choose this (simple) work-around is to safeguard that no new PHP 8.1/8.2 incompatibilities will be introduced in the plugin in the mean time while we're waiting for Core.
A "known to fail" build will be disregarded by most devs, which means that new incompatibilities can slip in a little too easily - as they have in Core -.

The test which is being skipped is a specific one, which _only_ tests for a specific PHP 8.1 deprecation to _not_ be thrown. It doesn't actually test the functionality in the plugin, which is why IMO this is the simplest way to safeguard the plugin for PHP 8.1/8.2 right now.
Also see: https://github.com/WordPress/wordpress-importer/pull/113

So yes, I agree it may not be the _best_ fix, but it is not meant to be. It is meant to only be a temporary measure,

jrfnl: WP 6.2 includes Requests 2.0.0, so when testing with WP 6.2, the tests should now pass without the first commit.

jrfnl:
Rebased the PR and added a second matrix update commit to allow for WP 6.2 and PHP 8.2 having been released.

jrfnl:
As the build is now passing, the first commit of this PR (what this PR was about) is no longer needed, so I'll close this PR.

I will open a new PR for the GH Actions matrix updates.

PR #140, opened 2022-11-30
Add wxr-parser-large-file.

vishnugopal:
This adds a more memory-efficient WXR file parser to the WordPress importer. The existing parser loads the entire WXR file into memory whereas this one uses XMLReader to parse the file in chunks.

This parser implementation has been extensively tested at WordPress.com, and reuses the existing parsers for maximum compatibility.

To test with large XML files [example](https://github.com/WordPress/wordpress-importer/files/10119518/vishnugopal.WordPress.2022-11-29.xml.zip):
- `wp-env start`
- `wp-env run cli vi .htaccess` 
- add: `php_value upload_max_filesize 256M`

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,0 +1,7 @@
+{
+  "plugins": ["."],
+  "config": {
+    "WP_UPLOAD_MAX_FILESIZE": "128M",
+    "WP_MEMORY_LIMIT": "256M"
+  }
+}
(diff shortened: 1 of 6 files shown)


.wp-env.json
```diff
@@ -0,0 +1,7 @@
+{
```
jeroenpf, line 1: You probably don't want to include this file :-)

jeroenpf (changes requested):
Delete the .wp-env.json file, or was it a conscious addition?

PR #143, opened 2023-03-15
Update paths to build status badges

danielbachhuber:


danielbachhuber:
@jrfnl The builds might not have started because I'm a "first time contributor". Can you see if you can get them started?

jrfnl:
> @jrfnl The builds might not have started because I'm a "first time contributor". Can you see if you can get them started?

I was looking at that, but that's not it. The workflows explicitly don't run when there are only changes to `txt` or `md` files, which is fine if there is no branch protection, but as there now is, I think that needs to be removed from the workflows.
Alternatively, committers should be allowed to bypass branch protection in those situations, but that's currently not allowed either.

I just checked, but I can't adjust the branch protection rules. @dd32 probably can, so let's hear their opinion on whether to allow bypassing or to adjust the workflows.

dd32:
> The workflows explicitly don't run when there are only changes to txt or md files, which is fine if there is no branch protection, but as there now is, I think that needs to be removed from the workflows.

I think removing those exclusions from the actions is the best route forward @jrfnl

> Alternatively, committers should be allowed to bypass branch protection in those situations, but that's currently not allowed either.

It looks like GitHub Admins and Repo Admins can bypass the restrictions, but having to bypass them doesn't seem like a good workflow in general.

Neither of you have explicit permissions on the repo, only Write access via the Core Committers group, that'll be why you can't alter things on the repo. Let me know if you need elevated access for anything.

jrfnl:
> > The workflows explicitly don't run when there are only changes to txt or md files, which is fine if there is no branch protection, but as there now is, I think that needs to be removed from the workflows.
> 
> I think removing those exclusions from the actions is the best route forward @jrfnl

See PR #145. Though it would also be good to merge my other related GH Actions PRs #136, #137, #138 and if we can resolve the discussion #131.

PR #146, opened 2023-04-07
Add support for WordPress 6.2

ouikhuan:
This PR adds support for WordPress 6.2 and upgrades the plugin to 0.9.

- Tested with WordPress 6.2
- Tested with PHP 5.6, PHP 7.4.33, PHP 8.0

How to test:

- Clone the plugin under your working directory
- Using `wp-env` to test, if you haven't installed it, please visit [wp-env](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#quick-tldr-instructions) for instructions
- It'll be easier if you have a `.wp-env.json` file at the root folder. For example, if your root folder is `oss-importer` and you can clone the plugin under this folder. Create `.wp.-env.json` file, you can change PHP version to the one you want to test with.
```
{
  "phpVersion": "5.6",
  "plugins": ["./wordpress-importer"]
}
``` 
- Run `wp-env start` to spin up the WordPress
- Navigate to `http://localhost:8888/wp-admin/admin.php?import=wordpress` and perform an import
- Import the content; it can take a while if you have a lot of content. If the script timeout, you need to set_time_limit to a higher value
- Check and see if the import works fine without errors

### DIFF
--- a/src/readme.txt
+++ b/src/readme.txt
@@ -3,9 +3,9 @@ Contributors: wordpressdotorg
 Donate link: https://wordpressfoundation.org/donate/
 Tags: importer, wordpress
 Requires at least: 5.2
-Tested up to: 6.1
+Tested up to: 6.2
 Requires PHP: 5.6
-Stable tag: 0.8
+Stable tag: 0.9
 License: GPLv2 or later
 License URI: https://www.gnu.org/licenses/gpl-2.0.html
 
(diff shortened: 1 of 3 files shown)


zaerl (approved):
LGTM!

jrfnl:
I must be missing something ? AFAICS this PR does not "add support" for WP 6.2. All this does is update the "tested up to" label.

There are no functional changes, so why should the version number be set to a new major release (minors below 1.0 are majors) - instead of `0.8.1` for instance ?

ouikhuan:
This is true, I'll change this with another PR and make it `0.8.1`

PR #148, opened 2023-04-07
Fix error workflow actions for assets

ouikhuan:
This PR fixes the error assets workflow by applying the develop version of `10up/action-wordpress-plugin-asset-update@develop`.

PR #149, opened 2023-04-07
Clean up readme.txt

ouikhuan:
Clean up readme.txt

### DIFF
--- a/src/readme.txt
+++ b/src/readme.txt
@@ -28,15 +28,15 @@ For further information and instructions please see the [documention on Importin
 The quickest method for installing the importer is:
 
 1. Visit Tools -> Import in the WordPress dashboard
-1. Click on the WordPress link in the list of importers
-1. Click "Install Now"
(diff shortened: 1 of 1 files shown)


zaerl (approved):
Nice find! 🚢

jrfnl:
This change is unnecessary. Markdown parses this to a numeric list and will fill in the numbers correctly as can be seen here: https://wordpress.org/plugins/wordpress-importer/#installation

Keeping all numbers at `1` prevents churn when a new list item would be added, so should be considered good practice.

ouikhuan:
> Keeping all numbers at 1 prevents churn when a new list item would be added, so should be considered good practice.

Thanks for pointing this out @jrfnl , we were trying to trigger the assets-readme.yml updates we made [here](https://github.com/WordPress/wordpress-importer/pull/148)  because we kept getting [fail build](https://github.com/WordPress/wordpress-importer/actions/runs/4636698907) through testing. In this case I'll close this PR.

PR #150, opened 2023-04-07
Update version to 0.8.1

ouikhuan:
Based on the suggestion [here](https://github.com/WordPress/wordpress-importer/pull/146#issuecomment-1500251181) it'll be better for us to use version 0.8.1 instead of 0.9

### DIFF
--- a/src/readme.txt
+++ b/src/readme.txt
@@ -5,7 +5,7 @@ Tags: importer, wordpress
 Requires at least: 5.2
 Tested up to: 6.2
 Requires PHP: 5.6
-Stable tag: 0.9
+Stable tag: 0.8.1
 License: GPLv2 or later
 License URI: https://www.gnu.org/licenses/gpl-2.0.html
 
@@ -40,7 +40,7 @@ If you would prefer to do things manually then follow these instructions:
 
 == Changelog ==
 
-= 0.9 =
+= 0.8.1 =
 
 * Update compatibility tested-up-to to WordPress 6.2.
 * Update paths to build status badges.
--- a/src/wordpress-importer.php
+++ b/src/wordpress-importer.php
@@ -6,7 +6,7 @@
  * Description:       Import posts, pages, comments, custom fields, categories, tags and more from a WordPress export file.
  * Author:            wordpressdotorg
  * Author URI:        https://wordpress.org/
- * Version:           0.9
+ * Version:           0.8.1
  * Requires at least: 5.2
  * Requires PHP:      5.6
  * Text Domain:       wordpress-importer
--- a/wordpress-importer.php
+++ b/wordpress-importer.php
@@ -1,7 +1,7 @@
 <?php
 /*
  * Plugin Name: WordPress Importer Git loader
- * Version: 0.9
+ * Version: 0.8.1
  */
 
 // This file is included purely for those using Git and checking out directly into wp-content/plugins/wordpress-importer/


jrfnl (approved):

jrfnl:
Side-note: version bumps are normally only done just before a release.

ouikhuan:
> Side-note: version number bumps are normally only done just before a release.

Since WordPress 6.2 was released earlier last week, we thought it would be a good idea to test the plugin and update it to the latest version of WordPress, and it seems we need a version update to trigger the sync from git repo to SVN? Sorry, we are relatively new to the workflow, what's the best action to do here if we want to update plugins that's tested up the latest WordPress?

Also, we kept getting [failure builds](https://github.com/WordPress/wordpress-importer/actions/runs/4638955180), is there any way we can fix this?

jrfnl:
> > Side-note: version number bumps are normally only done just before a release.
> 
> Since WordPress 6.2 was released earlier last week, we thought it would be a good idea to test the plugin and update it to the latest version of WordPress, and it seems we need a version update to trigger the sync from git repo to SVN? Sorry, we are relatively new to the workflow, what's the best action to do here if we want to update plugins that's tested up the latest WordPress?

The sync to WP is [only done on tags](https://github.com/WordPress/wordpress-importer/blob/876270f60f580e91a5bc3f32e4306149bbe3c481/.github/workflows/deploy.yml#L2-L5). To release a new version and sync with the WP SVN repo, a tag would need to be created. Updating the version number alone is not enough.

I think it might be a good idea to add a `CONTRIBUTING.md` and a release checklist to this repo with this kind of information. What do you think ?

If you need an example for what I mean by "release checklist" - here is one we use in the Requests project: https://github.com/WordPress/Requests/blob/develop/.github/release-checklist.md

> 
> Also, we kept getting [failure builds](https://github.com/WordPress/wordpress-importer/actions/runs/4638955180), is there any way we can fix this?

That workflow only updates the readme and plugin assets, it doesn't do a release.

I haven't looked at that particular failure yet. IIRC @dd32 added that workflow, so he may know more.

Based on a quick look at the log, the problem is "Other files have been modified; changes not deployed".
I would suggest checking in the repo of the action which is being used for the deploy if this has been reported before and what the recommendation to fix it was.
If it hasn't been reported before, I would suggest adding a "on failure" debug step to the workflow which would show which other files have been modified (maybe using something like `git status -b -v -u`), so at least you can pinpoint the problem.

Also note that it looks like a `develop` version of the [action runner](https://github.com/10up/action-wordpress-plugin-asset-update) is being used, so that could explain why things are unstable.
I presume that was so chosen as `develop`, at the time of adding the workflow, contained some fix which was needed for this repo, but it should be checked if it is possible to use a stable version of the action runner instead.

Having said that, it would be good to prevent other workflows from getting into problems _before they happen_, which is what this list of PRs is all about: https://github.com/WordPress/wordpress-importer/pulls/jrfnl
If I'm honest, those should be reviewed and merged _before_ anything else gets merged as CI is currently not up to date, so PRs/merges are not checked thoroughly enough as things are without merging those PRs.

ouikhuan:
> I think it might be a good idea to add a CONTRIBUTING.md and a release checklist to this repo with this kind of information. What do you think ?

Agreed! I think a checklist will be very helpful!

> Also note that it looks like a develop version of the [action runner](https://github.com/10up/action-wordpress-plugin-asset-update) is being used, so that could explain why things are unstable.

The reason why we tried to use a develop version is because @zaerl  from our team found a [bug](https://github.com/10up/action-wordpress-plugin-asset-update/commit/54e197027b1fbcf0a00c0b76790d8471df4666b4) previously but it's not landed in the stable version. We initially suspected that this might be causing our problems, so we attempted to switch to the develop branch to see if the build would succeed. However, even after making the switch, we were still unable to resolve the issue. If that's not the issue, maybe we can switch it back to the stable version?

@dd32 Could you suggest what's the best way moving forward? If the build is failure only because we modified the readme file and other files in the same PR, then I guess it should be fine to ignore it?

> Having said that, it would be good to prevent other workflows from getting into problems before they happen, which is what this list of PRs is all about: https://github.com/WordPress/wordpress-importer/pulls/jrfnl

Having checks to prevent problems from happening will definitely help IMHO

PR #151, opened 2023-04-28
Change assets workflow back to stable

ouikhuan:
Since the workflow wasn't the issue here, I'm changing it back to the stable version


zaerl (approved):
LGTM!

PR #155, opened 2023-12-27
Update tested up to label and test matrix

ouikhuan:
- This PR test for WordPress 6.4.2 with PHP 7.0, 7.4 and 8.0.
- Since latest WordPress is not supporting PHP 5.6 anymore, we also need to modify `test.yml`

How to test:

- Clone the plugin under your working directory
- Using `wp-env` to test, if you haven't installed it, please visit [wp-env](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#quick-tldr-instructions) for instructions
- It'll be easier if you have a `.wp-env.json` file at the root folder. For example, if your root folder is `oss-importer` and you can clone the plugin under this folder. Create `.wp.-env.json` file, you can change PHP version to the one you want to test with.
```
{
  "phpVersion": "7.0",
  "plugins": ["./wordpress-importer"]
}
``` 
- Run `wp-env start` to spin up the WordPress
- Navigate to `http://localhost:8888/wp-admin/admin.php?import=wordpress` and perform an import
- Import the content; it can take a while if you have a lot of content. If the script timeout, you need to set_time_limit to a higher value
- Check and see if the import works fine without errors

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,9 @@ jobs:
           - php: '7.0'
             wp: '5.8'
             experimental: false
+          - php: '7.0'
+            wp: '6.3'
+            experimental: false
 
           # Experimental builds. These are allowed to fail.
           - php: '8.1'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,10 +51,10 @@ jobs:
             wp: '5.3'
             experimental: false
           - php: '5.6'
-            wp: '5.2'
+            wp: '5.5'
             experimental: false
           - php: '5.6'
-            wp: '5.5'
+            wp: '5.2'
             experimental: false
           - php: '7.2'
             wp: '5.7'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         # Notes regarding supported versions in WP:
         # The base matrix only contains the PHP versions which are supported on all supported WP versions.
-        php: ['5.6', '7.0', '7.1', '7.2', '7.3']
+        php: ['7.0', '7.1', '7.2', '7.3']
         wp: ['latest', '5.2']
         experimental: [false]
 
@@ -49,6 +49,9 @@ jobs:
           # Complement the builds run via the matrix with some additional builds against specific WP versions.
           - php: '7.3'
             wp: '5.3'
+            experimental: false
+          - php: '5.6'
+            wp: '5.2'
             experimental: false
           - php: '5.6'
             wp: '5.5'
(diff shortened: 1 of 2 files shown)


ouikhuan:
I realized that even after removing PHP 5.6 from the `test.yml` matrix it's still showing up as expected check. Should we do this update in another PR?

.github/workflows/test.yml
```diff
@@ -53,6 +53,9 @@ jobs:
           - php: '5.6'
             wp: '5.5'
             experimental: false
+          - php: '5.6'
+            wp: '5.2'
+            experimental: false
```
jrfnl, line 58: WP only dropped support for PHP 5.6 in WP 6.3, so I'd suggest setting the `wp` key here to `6.3` to compliment the build against WP 5.5 which is already in the matrix.

jrfnl:
> I realized that even after removing PHP 5.6 from the `test.yml` matrix it's still showing up as expected check. Should we do this update in another PR?

That is unrelated to your PR. The required builds are defined in the protected branch settings and the way the branch protection is set up for this repo it will now literally block every single PR from being merged.

I just checked and while I can commit, I cannot change the branch protection, so you now have the not very enviable task to figure out who on this earth has admin rights for this repo and can set the branch protection to something a little more sane.

In my opinion, the following changes need to be made:
* The required checks need to be updated to be in sync with the current "non-experimental" builds. Quite a few are missing (and for this PR, one needs to be removed).
* The "Require branches to be up to date before merging " check needs to be removed as it is completely silly to require a rebase of all open PRs after each merge.
* The "Do not allow bypassing the above settings" check should be removed to allow select people to merge PRs even when the builds don't pass. This ability should generally only be used with discretion and careful after careful evaluation, but there are circumstances where multiple PRs need to be merged before a passing build can be achieved again and it is now impossible to get out of that type of situation.

Aside from the above, there are a number of PRs open (and rotting for over a year) which make various improvements to the GH Actions workflows. Those PRs should be merged with priority to keep things running smoothly and never be left to rot.

I would merge them myself now, except due to the long time between when I pulled them and now and having to rebase them due to the branch protection settings, I can't merge them anymore with the failing PHP 5.6 build.

In my opinion, we should also add a dependabot config to automatically update action runners as those are going out of date fast now too.

All in all, it comes down to nobody taking ownership of this repo and that being problematic when changes need to be made/merged/released.

ouikhuan:
Thanks for the suggestion @jrfnl , please let me know if I can help on those pending PRs from you. @vishnugopal, what do you think based on the suggestion above? Do we have permission to modify the settings?

.github/workflows/test.yml
ouikhuan: I only added in because previously wp 5.2 in the matrix was also checking php 5.6. But since WP latest, it's php 7 and above so php 5.6 needs to be removed from the matrix.

ouikhuan: Nice suggestion on adding wp 6.3 in for the check, I've added it now.

PR #157, opened 2024-01-04
Bump version up to 0.8.2

ouikhuan:
Bumping version up to include latest changes:

* Update compatibility tested-up-to to WordPress 6.4.2.
* Update doc URL references.
* Adjust workflow triggers.

### DIFF
--- a/src/readme.txt
+++ b/src/readme.txt
@@ -5,7 +5,7 @@ Tags: importer, wordpress
 Requires at least: 5.2
 Tested up to: 6.4.2
 Requires PHP: 5.6
-Stable tag: 0.8.1
+Stable tag: 0.8.2
 License: GPLv2 or later
 License URI: https://www.gnu.org/licenses/gpl-2.0.html
 
@@ -39,6 +39,12 @@ If you would prefer to do things manually then follow these instructions:
 1. Go to the Tools -> Import screen, click on WordPress
 
 == Changelog ==
+
+= 0.8.2 =
+
+* Update compatibility tested-up-to to WordPress 6.4.2.
+* Update doc URL references.
+* Adjust workflow triggers.
 
 = 0.8.1 =
 
--- a/src/wordpress-importer.php
+++ b/src/wordpress-importer.php
@@ -6,7 +6,7 @@
  * Description:       Import posts, pages, comments, custom fields, categories, tags and more from a WordPress export file.
  * Author:            wordpressdotorg
  * Author URI:        https://wordpress.org/
- * Version:           0.8.1
+ * Version:           0.8.2
  * Requires at least: 5.2
  * Requires PHP:      5.6
  * Text Domain:       wordpress-importer
--- a/wordpress-importer.php
+++ b/wordpress-importer.php
@@ -1,7 +1,7 @@
 <?php
 /*
  * Plugin Name: WordPress Importer Git loader
- * Version: 0.8.1
+ * Version: 0.8.2
  */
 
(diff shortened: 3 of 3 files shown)


src/readme.txt
```diff
@@ -40,6 +40,12 @@ If you would prefer to do things manually then follow these instructions:
 
 == Changelog ==
 
+= 0.8.2 =
+
+* Update compatibility tested-up-to to WordPress 6.4.2.
+* Update doc URL references.
+* Adjust workflow triggers.
+
```
jrfnl, line 48: None of these are user facing changes warranting a new release.

ouikhuan: That's correct, so we're releasing it with a patch version here. Otherwise, if old changes are merged and not shipped with a version bump, the changes like Tested up to and assets won't be able to porting over to the SVN directly in the future. Is there a better way to resolve this?

vishnugopal: When we first took up importer maintenance, several WordPress releases had passed by and the importer had an out of date notice on the plugin page. Verifying compatibility and releasing plugins with new tested up-to versions works well if we do it for every WordPress release, and that's what we'd like to do here.

jrfnl: > we're releasing it with a patch version here

A patch version is **_still_** a new release.

> the importer had an out of date notice on the plugin page

IIRC the "Tested up to" version on which that banner is based comes from the SVN `trunk` version of the plugin.

As long as the commit which updates the version is pushed to `trunk`, there is no need to release a new version to bump that header.

What you are doing is version inflation without benefit for end-users.

vishnugopal: I think the Github workflow for this repo won't allow us to commit to SVN directly without a release bump, but let's see what we can do to get around this.

jrfnl: @vishnugopal That would be a good reason to adjust/improve the workflow. Not a good reason for version inflation.

vishnugopal: I respectfully disagree. I think the benefit for end users here is that we're testing and verifying compatibility with the latest WordPress releases. But I'll defer to folks like @dd32 who have more experience releasing things like this.

dd32: > IIRC the "Tested up to" version on which that banner is based comes from the SVN trunk version of the plugin.

This isn't correct, unfortunately a full release is required to bump this field, which is my biggest gripe with how others (read: not-the-plugin-directory) treat the field.

Unfortunately due to "security" plugins, it's also become "good practice" to release a new version at least once a year, _even without any changes_, as otherwise they start complaining "you're using a possibly no longer supported plugin". Basically trying to justify their existence through fear generation.

While we've looked at adding a feature on the plugin directory to specify the tested-up-to without any code changes, unfortunately Core now includes things that use this (which I disagree with, and requires a plugin update to affect), which has also added an additional location for the header that we don't yet support https://meta.trac.wordpress.org/ticket/4621

https://github.com/WordPress/wordpress-develop/blob/24531a9afc8ef75822da45b85c147b492674e715/src/wp-admin/includes/class-wp-plugin-install-list-table.php#L813
> `echo '<span class="compatibility-untested">' . __( 'Untested with your version of WordPress' ) . '</span>';`

vishnugopal (approved):
Works with the latest WordPress version, approving!

src/readme.txt
vishnugopal: Thanks @dd32 based on your comments, we are going ahead and releasing this new version.

PR #158, opened 2024-01-04
GH Actions: update the matrix for latest WP/PHP

jrfnl:
This commit:
* Adds a proper set of builds against PHP 8.1, 8.2 and PHP 8.3 against low/high WP versions supported for those PHP versions as those PHP versions have been released and should be supported.
* Removes the "experimental" builds against PHP 8.1 and 8.2 as those PHP versions have been released and should be supported.
* Moves the builds against PHP 5.6 from the "complementary" builds section to the "PHP version support" section.
* Moves some comments around to make it more straight forward to understand how this matrix should be read and updated.

I've verified that all (nearly) WP versions between the lowest and the highest supported version now have at least one build.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,31 +24,55 @@ jobs:
         experimental: [false]
 
         include:
-          # Complement the builds run via the matrix with high/low WP builds for PHP 7.4 and 8.0.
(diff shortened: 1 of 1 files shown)


jrfnl:
Note: looks like the protected branch settings will need an update again... /cc @dion

dion:
don't @ me

Issue #162, opened 2024-04-09
Improve the post_exists query time during post import process.


dhusakovic:
During the post import we check whether the post already exists using [post_exists](https://developer.wordpress.org/reference/functions/post_exists/). 

The query currently utilises post title and date parameters, which will run through all the posts to try and find a match.

In cases where we have a large number of posts this query can time out and the import fail.

Passing the `post_type` to `post_exists` function allows us to use the `type_status_date` key, which significantly drops the number of rows examined and the overall query time in certain instances.

This change would also fix some edge case scenarios with posts across post types having the same title and date.

The update would be passing post_type argument to the check here:
https://github.com/WordPress/wordpress-importer/blob/master/src/class-wp-import.php#L659

So it would look like this
`$post_exists = post_exists( $post['post_title'], '', $post['post_date'], $post['post_type'] );`

dd32:
This looks like a safe change, a few lines later where `$post_exists` is used, the post_type is checked that it matches.. so querying only for that post_type shouldn't have any adverse effects.

dhusakovic:
Thanks for checking this out @dd32 🙇  
What's the process for creating a PR to address the issue? 
It looks like I don't have necessary access rights to push a branch up.

rebeccahum:
@dhusakovic You'll probably need to fork the repo

dd32:
Fixed by #163

PR #163, opened 2024-05-07
Update call to post_exists to include post_type in the query

dhusakovic:
During the post import we check whether the post already exists using [post_exists](https://developer.wordpress.org/reference/functions/post_exists/).

The query currently utilises post title and date parameters, which will run through all the posts to try and find a match.

In cases where we have a large number of posts this query can time out and the import fail.

Passing the `post_type` to `post_exists` function allows us to use the `type_status_date` key, which significantly drops the number of rows examined and the overall query time in certain instances.

This change would also fix some edge case scenarios with posts across post types having the same title and date.

Issue link: https://github.com/WordPress/wordpress-importer/issues/162

### DIFF
--- a/src/class-wp-import.php
+++ b/src/class-wp-import.php
@@ -656,7 +656,7 @@ class WP_Import extends WP_Importer {
 
 			$post_type_object = get_post_type_object( $post['post_type'] );
 
-			$post_exists = post_exists( $post['post_title'], '', $post['post_date'] );
+			$post_exists = post_exists( $post['post_title'], '', $post['post_date'], $post['post_type'] );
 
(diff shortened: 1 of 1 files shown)


dd32 (approved):
Per #162 this looks good to me.

The double-check on like 675 of `get_post_type( $post_exists ) == $post['post_type']` should probably stay too, as it runs after a filter of this value.

dhusakovic:
Hi @dd32 thanks for approving the PR! When can we expect this change to be merged?

rebeccahum:
@dd32 Will this make it into 6.6 you think?

PR #165, opened 2024-09-09
Add dependabot config

jrfnl:
I noticed a number of action runners used are out of date. Most updates are related to the Node version the action runners use under the hood updating to Node 20. Not having those updates will become problematic soonish when GHA drops support for Node 16.

So, instead of manually updating the workflows, I'm proposing enabling Dependabot to submit updates for the GHA runners. This should take care of it via Dependabot opening PRs to do the updates.


jrfnl:
Test failure is unrelated. See PR #166

PR #171, opened 2024-09-13
Start using WordPressCS 3.1

jrfnl:
#### Composer:
* No need for requiring the Composer PHPCS plugin anymore as it now comes automatically with WPCS and removing the requirement allows WPCS to manage the supported versions, preventing conflicts (and will update to the 1.0.0 version).

#### PHPCS ruleset:
* Minor tweak for a replaced sniff as per the changes in WPCS 3.0.0.
* Exclude one error code which would conflict with the minimum PHP requirement of this plugin.

Refs:
* https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/3.0.0
* https://github.com/WordPress/WordPress-Coding-Standards/wiki/Upgrade-Guide-to-WordPressCS-3.0.0-for-ruleset-maintainers

### CS/QA: all methods should have visibility declared

To prevent any BC-breaks, all methods without explicit visibility have now been explicitly declared as `public`.

Mind: the `set_up()` and `tear_down()` methods in the tests _should_ be `protected`, but as WP Core didn't always have this correct and this plugin still supports a large range of WP versions, so needs to take a large range of WP versions into account for the WP Core test framework on which the test suite is build, setting the visibility to `public` is the safer choice for now.

### CS/QA: all properties should have visibility declared

To prevent any BC-breaks, all properties without explicit visibility have now been explicitly declared as `public`.

### CS/QA: don't use reserved keywords as parameter names

... to prevent code getting really confusing when people would call it using PHP 8.0+ function calls with named arguments.

### CS/QA: miscellaneous other fixes

* Use pre-increment instead of post-increment.
* Use `require` for unconditional file includes.
* No blank line at end of function.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,7 @@
         "php": ">=5.6"
     },
     "require-dev": {
-        "wp-coding-standards/wpcs": "^2.3",
-        "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
+        "wp-coding-standards/wpcs": "^3.1",
         "phpcompatibility/phpcompatibility-wp": "^2.1",
         "yoast/phpunit-polyfills": "^1.0.1",
         "phpunit/phpunit": "^5.7.21 || ^6.5 || ^7.5"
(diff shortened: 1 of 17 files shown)


swissspidy (approved):
Appreciate the detailed changelog.

All sounds very reasonable to me 👍

PR #180, opened 2025-02-06
Remove SVN commands.

desrosj:
SVN has been deprecated and removed on the GitHub platform and is causing tests to fail. This switches the setup commands to use Git.

### DIFF
--- a/phpunit/install.sh
+++ b/phpunit/install.sh
@@ -26,7 +26,7 @@ else
 		echo "Latest WordPress version could not be found"
 		exit 1
 	fi
-	WP_TESTS_TAG="tags/$LATEST_VERSION"
+	WP_TESTS_TAG="$LATEST_VERSION"
 fi
 
 set -ex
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

jrfnl (commented):
The script which is being adjusted in this PR, looks to be an early/outdated version of the [`install-wp-tests.sh` script from the WP-CLI scaffold-command](https://github.com/wp-cli/scaffold-command/blob/main/templates/install-wp-tests.sh).

Would it make (more) sense to update the script with the latest version of the script from the wp-cli/scaffold-command repo instead ? 

If that would still cause issues with SVN, I suggest submitting a PR to the wp-cli/scaffold-command repo to update the script there before syncing the script used here with the upstream one. That way, many more people can benefit from any improvements.

jrfnl:
@desrosj Looks like you merged it while I was reviewing. Might be helpful if you ask people for a review to give them some time to do the review ?

PR #185, opened 2025-03-06
Add fix to XML meta values unserialize

zaerl:
This PR addresses a report and removes the `maybe_unserialize()` call made during the XML scan transform serialized classes to an `stdClass` and not other classes that can run code in the `__wakeup()` magic method.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
         experimental: [false]
 
         include:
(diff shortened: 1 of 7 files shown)


ouikhuan (changes requested):
Looks like we're missing one on line 975 in class-wp-import.php?

ouikhuan (approved):
LGTM

ouikhuan (approved):
I did a code review only and the approach makes sense to me.

PR #190, opened 2025-08-07
WXR_Parser_XML_Processor: An all-PHP WXR parser

adamziel:
## Description

Brings over the XMLProcessor class from [WordPress/php-toolkit](https://github.com/wordpress/php-toolkit) as a fallback XML parser. The end goal is removing the `WXR_Parser_Regex` class.

**This PR does not expose the new parser to the plugin consumers yet.** It only adds the new parser class and plugs it into the CI test pipeline.

## Rationale

`WXR_Parser_Regex` enables importing WXR files on hosts where libxml is not available. However, `WXR_Parser_Regex` cannot reliably parse XML documents. It does not conform to [the XML 1.0 spec](https://www.w3.org/TR/xml/) and fails multiple unit tests. `WXR_Parser_Regex` cannot be improved in small steps. Any iterative improvement strategy will ultimately lead to a full spec-compliant XML parser. That's what `XMLProcessor` is.

This PR proposes a new `WXR_Parser_XML_Processor` class that uses `XMLProcessor` as a backend. `XMLProcessor`:

* Understands XML nuances the regex-based parser can't process, e.g. namespaces, nesting, whitespaces, unicode.
* Is tested against the W3 XML test suite and is capable of stream-processing large XML files without running out of memory.
* Follows the ideas and conventions established in [WP_HTML_Processor](https://developer.wordpress.org/reference/classes/wp_html_processor/).

### Implementation details

This PR adds a copy of `WXREntityReader`, `XMLProcessor`, and their dependencies into this repository. This way:

* Cloning this repository is enough to use the `wordpress-importer` plugin, no `composer install` needed.
* `wordpress-importer` plugin remains small as it doesn't require pulling the entire `php-toolkit`.

In the future, it might be useful to switch to composer and find a way of pruning the installed dependencies.

### Follow-up work

* Measure the performance impact
* Bump PHP version checked by WPCS to 7.2 (same as latest WordPress core) to allow more static type annotations.
* Remove WXR_Parser_Regex.

cc @zaerl @dmsnell @sirreal @janjakes


src/parsers/xml-processor/class-xmlprocessor.php
```diff
@@ -0,0 +1,4499 @@
...
+ *
+ * @TODO: Track specific error states, expose informative messages, line
+ *        numbers, indexes, and other debugging info.
+ *
+ * @TODO: Skip over the following syntax elements:
+ *        * <!DOCTYPE, see https://www.w3.org/TR/xml/#sec-prolog-dtd
+ *        * <!ATTLIST, see https://www.w3.org/TR/xml/#attdecls
+ *        * <!ENTITY, see https://www.w3.org/TR/xml/#sec-entity-decl
+ *        * <!NOTATION, see https://www.w3.org/TR/xml/#sec-entity-decl
+ *        * Conditional sections, see https://www.w3.org/TR/xml/#sec-condition-sect
```
dmsnell, line 43: skip over or bail? I guess if we ignore we can bail on an entity whose definition we don’t have.

src/parsers/xml-processor/class-xmlprocessor.php
```diff
@@ -0,0 +1,4499 @@
...
+ *     $processor = XMLProcessor::create_from_string( '<style>// there could be more styling to come' );
+ *     false === $processor->next_tag();
+ *
+ *     $processor = XMLProcessor::create_from_string( '<style>// this is everything</style><content>' );
+ *     true === $processor->next_tag( 'content' );
+ *
+ * #### Special elements
+ *
+ * All XML elements are handled in the same way, except when you mark
+ * them as PCData elements. These are special because their contents
```
dmsnell, line 193: I think I’ve been wrong about this before on two accounts:

 - `PCDATA` is the normative text type we’re used to working with in HTML. tags inside PCDATA content are indeed tags, not plaintext.
 - `CDATA` is the verbatim text type, but an element cannot specify its content model as CDATA. an attribute’s content may be CDATA, but not an element’s.

this clears up the long confusion I had around differentiating content with a text span that looks like the closing tag of the wrapping CDATA element — it can’t happen. the only way to embed CDATA in an element is inside a CDATA section with `<![CDATA[…]]>`.

now I think we have some other concerns here with things like un-closed `post-content` elements. I’m directly inspired by the HTML parser’s stack machine for making decisions like “if there is an open `post-content` on the stack of open elements, generate implied end tags; otherwise ignore”

we might want or need this in the design, or announce it in a dry-run

src/parsers/xml-processor/utf8.php
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	// Any values above U+10FFFF are eliminated above in the pre-check.
+	$byte1 = chr( ( $code_point >> 18 ) | 0xF0 );
+	$byte2 = chr( ( $code_point >> 12 ) & 0x3F | 0x80 );
+	$byte3 = chr( ( $code_point >> 6 ) & 0x3F | 0x80 );
+	$byte4 = chr( $code_point & 0x3F | 0x80 );
+
+	return "{$byte1}{$byte2}{$byte3}{$byte4}";
+}
```
dmsnell, line 363: resurrecting in [Core-63863](https://core.trac.wordpress.org/ticket/63863)

src/parsers/xml-processor/class-wxrentityreader.php
```diff
@@ -0,0 +1,984 @@
...
+ * only reads meaningful entities from the WXR file.
+ *
+ * ## Design goals
+ *
+ * WP_WXR_Entity_Reader is built with the following characteristics in mind:
+ *
+ * * Speed – it should be as fast as possible
+ * * No PHP extensions required – it can run on any PHP installation
+ * * Reliability – no random crashes when encountering malformed XML or UTF-8 sequences
+ * * Low, predictable memory footprint to support 1000GB+ WXR files
```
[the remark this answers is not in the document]
adamziel: I was shocked but they actually exist!

src/parsers/xml-processor/class-xmlprocessor.php
```diff
@@ -0,0 +1,4499 @@
...
+ * * UTF-8 encoded
+ * * Not standalone (so can use external entities)
+ * * No DTD, DOCTYPE, ATTLIST, ENTITY, or conditional sections
+ *
+ * ### Possible future direction for this module
+ *
+ * The final goal is to support both 1.0 and 1.1 depending on the
+ * initial processing instruction (<?xml version="1.0" ?>). We're
+ * starting with 1.0, however, because most that's what most WXR
+ * files declare.
```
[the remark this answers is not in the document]
adamziel: Aha – really good idea! I'm applying these changes in the wordpress-components repo and will ping you there when I have a PR

src/parsers/xml-processor/class-xmlprocessor.php
adamziel: We bail out on them. To ignore them, we'd need to be able to parse them. I'll update the comment to specify this more precisely.

src/parsers/xml-processor/class-xmlprocessor.php
adamziel: Aha! This is really useful, thank you! I'll remove this entire semantics, then.

> now I think we have some other concerns here with things like un-closed post-content elements. 

No WXR importer can process that today, right? I'd love to eventually introduce an error token to put the API consumer in charge of what to do, though.

src/parsers/xml-processor/class-wxrentityreader.php
```diff
@@ -0,0 +1,984 @@
...
+ *        </wp:postmeta>
+ *        <wp:post_id>146</wp:post_id>
+ *      </item>
+ *  </channel>
+ * </rss>
+ * ```
+ *
+ * `WP_WXR_Entity_Reader` would accumulate post data until the `wp:post_meta` tag. Then it would emit a
+ * `post` entity and accumulate the meta information until the `</wp:postmeta>` closer. Then it
+ * would advance to `<wp:post_id>` and **ignore it**.
```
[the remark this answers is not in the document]
adamziel: We have `$xml_processor->bail()` that's inspired by the HTML API:

```php
			$this->bail(
				'Invalid closing tag encountered.',
				self::ERROR_SYNTAX
			);
```

It stores some contextual information that can be used later on to create a nice error message:

```php
		$this->exception  = new XMLUnsupportedException(
			$message,
			$this->get_token_type(),
			$starts_at,
			$token,
			$this->get_breadcrumbs()
		);
```

Now, your comment is more about higher-order concepts in the WXR file, like unrecognized entities. We don't have a standard way of announcing these at the moment. In the context of the `wordpress-importer` plugin, we'd need something that works with all the XML parsers shipped in this repo – `WXREntityReader` is specific to `XMLProcessor`.

One way to get there would be to update every WXR parser in here to provide meaningful error details in every error scenario. I'm nervous about this design as it requires consistent and correct reporting of the same errors across multiple parser implementations. I'm worried they'd diverge over time.

A safer way would be to rewire all the data flows in this repo from:

```
WXR_Parser_SimpleXML -> WP_Import
WXR_ParserXML -> WP_Import
XMLProcessor -> WXREntityReader -> WXR_Parser_XML_Processor -> WP_Import
```

To:

```
WXR_Parser_SimpleXML -> WXREntityReader -> WP_Import
WXR_ParserXML -> WXREntityReader -> WP_Import
XMLProcessor -> WXREntityReader -> WXR_Parser_XML_Processor -> WP_Import
```

And detect the errors at the WXREntityReader level. The XML parsing classes would be reduced to a consistent interface for walking through an XML document.

adamziel:
@dmsnell thank you for your great review! I've addressed most of your points in https://github.com/WordPress/php-toolkit/pull/149 and the rest of them inline.

src/parsers/xml-processor/class-wxrentityreader.php
```diff
@@ -0,0 +1,984 @@
...
+				'category' === $this->xml->get_tag_local_name() &&
+				array_key_exists( 'domain', $this->last_opener_attributes ) &&
+				array_key_exists( 'nicename', $this->last_opener_attributes )
+			) {
+				$this->entity_data['terms'][] = array(
+					'taxonomy'    => $this->last_opener_attributes['domain'],
+					'slug'        => $this->last_opener_attributes['nicename'],
+					'description' => $this->text_buffer,
+				);
+				$this->text_buffer            = '';
```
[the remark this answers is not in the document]
adamziel: Thank you for bringing this up @dmsnell! I think we need to leave this up to the XMLProcesor consumer. Here's my reasoning:

ICU with it's Normalizer class is not always available and I'd hate to output different bytes for the same document in different runtimes.

[XML 1.0 spec](https://www.w3.org/TR/xml/) doesn't seem to be making any normalization recommendations. [XML 1.1](https://www.w3.org/TR/xml-c14n/), however, discusses normalization in section 4.2 No Character Model Normalization:

> The Unicode standard [[Unicode]](https://www.w3.org/TR/xml-c14n/#Unicode) allows multiple different representations of certain "precomposed characters" (a simple example is "ç"). Thus two XML documents with content that is equivalent for the purposes of most applications may contain differing character sequences. The W3C is preparing a normalized representation [[CharModel]](https://www.w3.org/TR/xml-c14n/#CharModel). The [C14N-20000119](https://www.w3.org/TR/xml-c14n/#C14N-20000119) Canonical XML draft used this normalized form. However, many **XML 1.0 processors do not perform this normalization**. Furthermore, applications that must solve this problem typically enforce character model normalization at all times starting when character content is created in order to avoid processing failures that could otherwise result (e.g. see example from [Cowan](https://www.w3.org/TR/xml-c14n/#CowanExample)). Therefore, character model normalization has been moved out of scope for XML canonicalization. However, the XML processor used to prepare the XPath data model input is required (by the [Data Model](https://www.w3.org/TR/xml-c14n/#DataModel)) to use Normalization Form C [[NFC](https://www.w3.org/TR/xml-c14n/#ref-NFC), [NFC-Corrigendum](https://www.w3.org/TR/xml-c14n/#NFC-Corrigendum)] when converting an XML document to the UCS character domain from any encoding that is not UCS-based (currently, UCS-based encodings include UTF-8, UTF-16, UTF-16BE, and UTF-16LE, UCS-2, and UCS-4).

The entity reader could be making note of unique normalized category names whenever ICU is available and making a recommendation to merge them.

src/parsers/xml-processor/class-basebytereadstream.php
```diff
@@ -0,0 +1,186 @@
...
+		return substr( $this->buffer, $this->offset_in_current_buffer, $n );
+	}
+
+	public function consume( $n ) {
+		if ( strlen( $this->buffer ) < $this->offset_in_current_buffer + $n ) {
+			throw new NotEnoughDataException( 'Cannot consume more bytes than available in the buffer.' );
+		}
+		$bytes                           = substr( $this->buffer, $this->offset_in_current_buffer, $n );
+		$this->offset_in_current_buffer += $n;
+		if ( $this->offset_in_current_buffer > $this->buffer_size ) {
```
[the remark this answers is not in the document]
adamziel: Ditto the other comment about `$buffer_size`. It's now called `$max_lookbehind_bytes` and it only means the number of bytes we keep around to `seek()` back when needed – 2048 is typically enough and it can be adjusted for the scenario when more is needed.

src/parsers/xml-processor/class-xmlprocessor.php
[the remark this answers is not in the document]
adamziel: Addressed in https://github.com/WordPress/php-toolkit/pull/150 and backported here

src/parsers/xml-processor/class-xmlprocessor.php
adamziel: I've updated the comment

src/parsers/class-wxr-parser-xml-processor.php
```diff
@@ -0,0 +1,372 @@
...
+		// Every XML element is a combination of a long-form namespace and a
+		// local element name, e.g. a syntax <wp:post_id> could actually refer
+		// to a (https://wordpress.org/export/1.0/, post_id) element.
+		//
+		// Namespaces are paramount for parsing XML and cannot be ignored. Elements
+		// element must be matched based on both their namespace and local name.
+		//
+		// Unfortunately, different WXR files defined the `wp` namespace in a different way.
+		// Folks use a mixture of HTTP vs HTTPS protocols and version numbers. We must
+		// account for all possible options to parse these documents correctly.
```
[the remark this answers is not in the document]
adamziel: > I’m guessing there are subtle claimed differences between the given version numbers…

I'm guessing so too, they can probably be reverse-engineered by studying previous plugin releases.

adamziel:
@zaerl I'm really interested in your take on this PR

src/parsers/xml-processor/utf8.php
```diff
@@ -0,0 +1,363 @@
...
+	) {
+		return '�';
+	}
+
+	if ( $code_point <= 0x7F ) {
+		return chr( $code_point );
+	}
+
+	if ( $code_point <= 0x7FF ) {
+		$byte1 = chr( ( $code_point >> 6 ) | 0xC0 );
```
zaerl, line 342: This is similar to [what I do](https://github.com/zaerl/mojibake/blob/main/src/encoding.c#L196), and others as well. What we have here is working because we are checking U+0080–U+07FF.

The mask I use is still considered a good practice: it documents that you want to “keep 5 bits” and not the rest of the integer. Other checks that follow are the same.

src/parsers/xml-processor/class-importentity.php
```diff
@@ -0,0 +1,66 @@
...
+		'comment_status',
+		'ping_status',
+		'post_name',
+		'post_parent',
+		'menu_order',
+		'post_type',
+		'post_password',
+		'is_sticky',
+		'attachment_url',
+	);
```
zaerl, line 45: I'm not very sure about this list here. Where did we get this? Shouldn't it be the list found on `WXREntityReader::__construct`?

src/parsers/xml-processor/class-wxrentityreader.php
```diff
@@ -0,0 +1,984 @@
...
+		// Unfortunately, different WXR files defined the `wp` namespace in a different way.
+		// Folks use a mixture of HTTP vs HTTPS protocols and version numbers. We must
+		// account for all possible options to parse these documents correctly.
+		$wxr_namespaces       = array(
+			'http://wordpress.org/export/1.0/',
+			'https://wordpress.org/export/1.0/',
+			'http://wordpress.org/export/1.1/',
+			'https://wordpress.org/export/1.1/',
+			'http://wordpress.org/export/1.2/',
+			'https://wordpress.org/export/1.2/',
```
zaerl, line 327: I like the approach here, even if it generates `x6` more data. Great work.

src/parsers/xml-processor/utf8.php
```diff
@@ -0,0 +1,363 @@
...
+ * @param  string  $bytes  Text for which to count code points.
+ * @param  int|null  $first_error_byte_at  Optional. If provided, will be set upon finding
+ *                                      the first invalid byte.
+ *
+ * @return int How many code points were decoded in the given byte stream before an error
+ *             or before reaching the end of the string.
+ * @since {WP_VERSION}
+ *
+ */
+function utf8_code_point_count( $bytes, &$first_error_byte_at = null ) {
```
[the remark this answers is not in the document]
adamziel: Renamed, thank you

src/parsers/xml-processor/class-xmlunsupportedexception.php
```diff
@@ -0,0 +1,97 @@
...
+
+use Exception;
+
+/**
+ * Core class used by the XML processor during XML parsing
+ * for indicating that a given operation is unsupported.
+ *
+ * This class is designed for internal use by the XML processor.
+ *
+ * The XML API aims to operate in compliance with the XML5
```
zaerl, line 16: Did you mean XML 1.1?

adamziel: Ah, I meant XML 1.0. Not sure where 5 came from

zaerl:
I really like this, some fresh air is much needed in our (legacy?) WordPress importer plugin. We can land this.

src/parsers/xml-processor/utf8.php
```diff
@@ -0,0 +1,363 @@
...
+		"\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09\x09" .
+		"\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07" .
+		"\x08\x08\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02\x02" .
+		"\x10\x03\x03\x03\x03\x03\x03\x03\x03\x03\x03\x03\x03\x04\x03\x03" .
+		"\x11\x06\x06\x06\x05\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08" .
+		"\x00\x01\x02\x03\x05\x08\x07\x01\x01\x01\x04\x06\x01\x01\x01\x01" .
+		"\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x00\x01\x01\x01\x01\x01\x00\x01\x00\x01\x01\x01\x01\x01\x01" .
+		"\x01\x02\x01\x01\x01\x01\x01\x02\x01\x02\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x02\x01\x01\x01\x01\x01\x01\x01\x01" .
+		"\x01\x02\x01\x01\x01\x01\x01\x01\x01\x02\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x03\x01\x03\x01\x01\x01\x01\x01\x01" .
+		"\x01\x03\x01\x01\x01\x01\x01\x03\x01\x03\x01\x01\x01\x01\x01\x01\x01\x03\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01"
```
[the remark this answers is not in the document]
adamziel: I think @dmsnell is tweaking that entire code in https://core.trac.wordpress.org/ticket/63863 – I'll leave this table as it is for now and then backport any improvements Dennis makes in his patch.

zaerl (approved):

src/parsers/xml-processor/class-importentity.php
adamziel: I don't think this constant is even used in this PR, I'll remove it

src/parsers/xml-processor/utf8.php
```diff
@@ -0,0 +1,363 @@
...
+	$type      = ord( $state_table[ $byte ] );
+	$codepoint = ( UTF8_DECODER_ACCEPT === $state )
+		? ( ( 0xFF >> $type ) & $byte )
+		: ( ( $byte & 0x3F ) | ( $codepoint << 6 ) );
+
+	return ord( $state_table[ 256 + ( $state * 16 ) + $type ] );
+}
+
+/**
+ * Extract a slice of a text by code point, where invalid byte seuqences count
```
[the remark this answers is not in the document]
adamziel: Reworded, thanks

PR #192, opened 2025-08-08
Unblock CI runs

adamziel:
Renames the `test.yml` CI job, that for some reason got stuck for all the PRs, into `tests.yml`, which somehow runs in GitHub CI without any issues.

Before: 

CleanShot 2025-08-08 at 12 56 19@2x

After:

(tbd)


adamziel:
> This should never have been merged in the first place.

1. Why? Does it have any downsides?
2. I'll take that as "yes, I'd like to revert" and open a new PR.

jrfnl:
> [git blame](https://git-scm.com/docs/git-blame) doc states:
> 
> > The origin of lines is automatically followed across whole-file renames
> 
> I confirmed in my CLI the `blame` output is still in tact:
> 
> So `blame` history isn't an issue. Are there any other potential problems with the rename?

It may look that way now with, but it rarely is in reality. The break won't be visible until there are new commits to the renamed file.

PR #196, opened 2025-09-03
Bump minimum PHP version to 7.2

adamziel:
WordPress no longer supports PHP 5.6, 7.0, and 7.1 This PR updates the `wordpress-importer` plugin to reflect that state of affairs. To run on older PHP versions, a legacy version of the plugin can still be used.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
         experimental: [false]
 
         include:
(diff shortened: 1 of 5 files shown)


desrosj:
Just a heads up that this caused the Core PHPUnit test workflows in branches 6.7-5.2 to break. The plugin used to be cloned and tested within the Core unit test suite until r59769, and these old branches when running on PHP < 7.2.

It's resolved and addressed in Core-63983.

PR #207, opened 2025-09-12
Enable or disable URL rewriting with a checkbox

adamziel:
Now that we have the ability to [rewrite URLs](https://github.com/WordPress/wordpress-importer/pull/202) in the imported content, let's expose it to the users.

This PR adds a checkbox in the author mapping step that sets URL rewriting on or off. It is checked by default:

CleanShot 2025-09-12 at 14 36 03@2x

CC @schlessera @swissspidy – you might want to add a WP-CLI flag to enable the `rewrite_urls` option. Also cc @zaerl @brandonpayton

## Testing instructions

CI. E2E tests cover a scenario where the checkbox is checked and unchecked.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -8,7 +8,12 @@ on:
 
 jobs:
   e2e:
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3]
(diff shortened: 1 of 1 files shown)


swissspidy (approved):
Haven't tested this at all. Purely looked at the code changes.

PR #209, opened 2025-09-16
Add a version bump script

adamziel:
Adds a `composer bump-version` script that automatically bumps all the relevant version numbers. Also adds a CI job to run it from GitHub UI.


adamziel:
Closing this PR. Sadly, it won't be of big help as the changelog still needs manual curation if we don't want to include technical entries such as "version bump" etc.

PR #213, opened 2025-09-18
Rewrite relative URLs in block attributes

adamziel:
Backports https://github.com/WordPress/php-toolkit/pull/188 and adds an e2e test to ensure that relative URLs in block attributes are correctly detected and migrated.


zaerl (approved):
Already saw the original PR, and it was great. Thanks for solving these nitpicks here. Having full coverage is the best base from which to build optimizations.

Issue #217, opened 2025-09-23
Importer breaks image paths when Media Settings differ between export and import sites


bkjproductions:
**Description**

When importing an XML export file from one WordPress site into another, image paths break if the source and destination sites have different **Media Settings** → “Organize my uploads into month- and year-based folders.”

---

**Steps to reproduce**

**On Site A:**
- **Disable** “Organize my uploads into month- and year-based folders.”
- Upload a few images (they are stored **flat** in `/wp-content/uploads/`).
- Create posts referencing those images.
- Export content (Tools → Export).

**On Site B:**
- **Enable** “Organize my uploads into month- and year-based folders.”
- Import the XML export from Site A.
- Observe the imported posts.

---

**Actual behavior**
- The importer downloads the image files and saves them into `/uploads/YYYY/MM/` folders (according to Site B’s settings).
- Post content and attachment metadata still point to `/uploads/filename.jpg` (the original flat structure).
- **Result:** broken image links in all posts.

---

**Expected behavior**
- Importer should either:
  - Preserve the original upload paths from the export file, **or**
  - At minimum, **warn** the user if Media Settings differ between source and destination.

---

**Impact**
- Common during site migrations.  
- Results in broken media across large sites (hundreds of posts).  
- Requires writing custom scripts or server rewrites to fix paths.  

---

**Suggested fixes**

- Add a setting during import:  
  - [ ] Preserve original media paths (recommended for migrations)  
  - [ ] Reorganize into current Media Settings (default)  

- Or, add a clear warning:  
  _“Your current Media Settings differ from the export site. Imported images may not display correctly.”_  

- Or, store the Media Settings **within** the XML file so this would be easy to flag before import.  
  There is a `wp-options` setting: `uploads_uses_yearmonth_folders`, so we could just add to the `<channel>` node:
```xml
<wp:uploads_uses_yearmonth_folders>1</wp:uploads_uses_yearmonth_folders>
```

I realize, the user should perhaps configure the destination site to use the **same** Media Library settings, but the point of a tool is to save work, and this could save people hours of cleanup. 

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/51f17184-eccd-4d5b-84b5-8382d5d0f679" />

adamziel:
https://github.com/WordPress/wordpress-importer/pull/220 should fix it, @zaerl would you help with reviews?

adamziel:
[...]

PR #220, opened 2025-09-25
Rewrite attachment URLs to the new URL structure

adamziel:
Ensures the correct remapping of URLs referencing media files found in the imported posts when URL rewriting is enabled.

Solves https://github.com/WordPress/wordpress-importer/issues/217

### Problem

When URL rewriting is enabled, posts such as this one:

```html
<img src="https://my-old-site.com/subpath/wp-content/uploads/2008/06/canola2.jpg" />
```

Are imported as:

```html
<img src="https://my-new-site.com/wp-content/uploads/2008/06/canola2.jpg" />
```

The media URL rewriting, however, tries to do a second pass over the imported content to replace the original media URL with the current site's media URL. Assuming we're importing from a `YYYY/MM/{filename}` uploads structure to a flat `{filename}` structure, that media rewriting step would try to replace

```
https://my-old-site.com/subpath/wp-content/uploads/2008/06/canola2.jpg
```

with 

```
https://my-new-site.com/wp-content/uploads/canola2.jpg
```

However, that initial URL no longer exists in the content as the base URL was already migrated.

### Solution

This PR migrates the base URL in the media-related `url_remap` entries to ensure we're replacing the URL that were effectively inserted into the database during the import. For example, it migrates `https://my-old-site.com/subpath/wp-content/uploads/2008/06/canola2.jpg` to `https://my-new-site.com/wp-content/uploads/2008/06/canola2.jpg` before creating a `url_remap` entry to migrate it to `https://my-new-site.com/wp-content/uploads/canola2.jpg`.

cc @zaerl

### DIFF
--- a/src/class-wp-import.php
+++ b/src/class-wp-import.php
@@ -99,6 +99,11 @@ class WP_Import extends WP_Importer {
 
 		$this->options = apply_filters( 'wp_import_options', $options );
 
+		add_filter( 'import_post_meta_key', array( $this, 'is_valid_meta_key' ) );
+		add_filter( 'http_request_timeout', array( &$this, 'bump_request_timeout' ) );
+
+		$this->import_start( $file );
+
 		/**
 		 * If URL rewriting was requested but the WP version is too old, report
 		 * an error and disable it.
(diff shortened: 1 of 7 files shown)


src/php-toolkit/DataLiberation/URL/class-wpurl.php
```diff
@@ -25,6 +25,118 @@ public static function can_parse( $url, $base = null ) {
...
+				$from_pathname,
+				strlen( $base_pathname_with_trailing_slash )
+			);
+			$to_pathname_with_trailing_slash   = rtrim( $to_pathname, '/' ) . '/';
+			$remaining_pathname                = substr(
+				$decoded_matched_pathname,
+				strlen( $base_pathname_with_trailing_slash )
+			);
+
+			$updated_url->pathname = $to_pathname_with_trailing_slash . $remaining_pathname;
```
zaerl, line 86: 👌

src/class-wp-import.php
```diff
@@ -99,6 +99,11 @@ public function import( $file, $options = array() ) {
 
 		$this->options = apply_filters( 'wp_import_options', $options );
 
+		add_filter( 'import_post_meta_key', array( $this, 'is_valid_meta_key' ) );
+		add_filter( 'http_request_timeout', array( &$this, 'bump_request_timeout' ) );
```
zaerl, line 103: Let's `remove_filter`, to be good neighbors.

zaerl (approved):
Great work here. I've tested it and it's a slick solution. I've added a couple of suggestions, but we can ignore them now. Nothing critical.

src/class-wp-import.php
adamziel: This merely moves the add_filter call from below so I'll leave `remove_filter` out of this PR

PR #224, opened 2025-09-30
Fix PHP 8.3 compatibility warnings in regex parser

### LINKED ISSUE #176: Not php 8.3 compatible?
When i import the wordpress demo data i get these numerous warnings on the same line.

Warning: Undefined array key 1 in /var/docs/public/wp-content/plugins/wordpress-importer/parsers/class-wxr-parser-regex.php on line 65

Deprecated: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /var/docs/public/wp-content/plugins/wordpress-importer/parsers/class-wxr-parser-regex.php on line 118

Please resolve

I have debug mode enabled.  A clean wordpress install with a default wp theme.

I am importing https://github.com/WordPress/theme-test-data file themeunittestdata.wordpress.xml

prab18hat:
- Add isset() check before accessing array key in author processing (line 65)
- Add null check before preg_match() to prevent deprecated warning (line 118)
- Add similar safety checks for other array access patterns
- Maintains backward compatibility while eliminating PHP 8.3 warnings

Fixes #176

### DIFF
--- a/src/parsers/class-wxr-parser-regex.php
+++ b/src/parsers/class-wxr-parser-regex.php
@@ -43,18 +43,18 @@ class WXR_Parser_Regex {
 				$importline  = rtrim( $this->fgets( $fp ) );
 
 				if ( ! $wxr_version && preg_match( '|<wp:wxr_version>(\d+\.\d+)</wp:wxr_version>|', $importline, $version ) ) {
-					$wxr_version = $version[1];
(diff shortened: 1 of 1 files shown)


BackuPs:
If i am not mistaken: Support for the Regex Parser is abandoned as fas as i understood.  The new version does not call it anymore.  It is only there for backwards compatibility for programmers who are still calling it.

PR #248, opened 2025-11-29
Prefix vendor namespaces to prevent name conflicts

### LINKED ISSUE #245: Request: Prefixed/Namespaced Vendor Packages to Prevent Dependency Conflicts (e.g., psr/, symfony/, brick/math)
A similar issue has been discussed here:
🔗 [wp-cli/wp-cli#5920](https://github.com/wp-cli/wp-cli/issues/5920)

To avoid such conflicts, it would be extremely helpful if the vendor dependencies were **prefixed or namespaced**, similar to how code is **downgraded using Rector**. This approach would isolate the bundled packages and ensure conflict-free distribution.

[`/src/php-toolkit/DataLiberation/vendor-patched`](https://github.com/WordPress/wordpress-importer/tree/master/src/php-toolkit/DataLiberation/vendor-patched)

Initially, I assumed that the `vendor-patched` directory contained **prefixed dependencies** as well. However, it appears that these libraries have only been **downgraded to PHP 7.2 compatibility** using Rector and some manual fixes.

Extend the existing “vendor-patched” process to also include **namespace scoping/prefixing** of bundled dependencies to prevent conflicts.

adamziel:
## Summary

When the WordPress Importer is used alongside other plugins or WP-CLI packages that bundle the same libraries (like `rowbot/url`, `brick/math`, or `psr/log`), PHP class redefinition errors occur because both distributions load identically-named classes.

This change adds a `VendorPrefix\` prefix to all bundled vendor namespaces, so classes like `Rowbot\URL\URL` become `VendorPrefix\Rowbot\URL\URL`. This isolates our dependencies and allows the importer to coexist with any other package that bundles the same libraries.

 ## Test plan

Confirm all the CI checks are green.

Fixes #245


zaerl (approved):
This is a good solution. Not checking all the ~200 files because I'm sure there is no missing part, and a green status is enough for me.

Log1x:
@adamziel really appreciate this change! Wondering if we could get a release with this? The latest zip from .org still has dependencies like `Psr\Log` with no namespace prefix in place. Thank you!

PR #251, opened 2026-01-09
Prevent URL remapping for attachments without file extensions

### LINKED ISSUE #186: Remapping once media import does not works as expected for media hosted externally without file extension
## Issue

- For example, create a WXR file with minimum of requirements `wp_attachment_url` with an external URL without extension. 
- For example, create a new Google Doc add an image to the doc and use the static URL provided, it would be something like https://lh7-rt.googleusercontent.com/docsz/AD_4nXepU93bn7nmRtZwpqQOGO2PeIwurdNY1DCOl6WBIrPnYwLFOz1DfoLZD8hcHY2Es7LYQBUIxqB2aHl3o89Xlg2RWwnG2NLdA7NAYm57IhyzI6uvhIMZx4grzgUsCShkZb5VrqFV?key=KJq1w2dlYSfo8kldshcGURv7

XML file format,
```
<item>
		<title><![CDATA[view-after-2]]></title>
		<dc:creator><![CDATA[admin]]></dc:creator>
		<wp:post_name><![CDATA[view-after-2]]></wp:post_name>
		<wp:status><![CDATA[inherit]]></wp:status>
		<wp:post_type><![CDATA[attachment]]></wp:post_type>
		<wp:attachment_url><![CDATA[https://lh7-rt.googleusercontent.com/docsz/AD_4nXepU93bn7nmRtZwpqQOGO2PeIwurdNY1DCOl6WBIrPnYwLFOz1DfoLZD8hcHY2Es7LYQBUIxqB2aHl3o89Xlg2RWwnG2NLdA7NAYm57IhyzI6uvhIMZx4grzgUsCShkZb5VrqFV?key=KJq1w2dlYSfo8kldshcGURv7]]></wp:attachment_url>
</item>
```

Add this item to the WXR file for the post, and use the image block with source as the link given.

```
<item>
		<title><![CDATA[Image with External URL]]></title>
		<dc:creator><![CDATA[admin]]></dc:creator>
		<content:encoded><![CDATA[<!-- wp:image {"sizeSlug":"large"} -->
<figure class="wp-block-image size-large"><img src="https://lh7-rt.googleusercontent.com/docsz/AD_4nXepU93bn7nmRtZwpqQOGO2PeIwurdNY1DCOl6WBIrPnYwLFOz1DfoLZD8hcHY2Es7LYQBUIxqB2aHl3o89Xlg2RWwnG2NLdA7NAYm57IhyzI6uvhIMZx4grzgUsCShkZb5VrqFV?key=KJq1w2dlYSfo8kldshcGURv7" alt=""/></figure>
<!-- /wp:image -->]]></content:encoded>
		<wp:post_date><![CDATA[2025-03-24 11:08:36]]></wp:post_date>
		<wp:post_date_gmt><![CDATA[2025-03-24 11:08:36]]></wp:post_date_gmt>
		<wp:post_modified><![CDATA[2025-03-24 12:08:44]]></wp:post_modified>
		<wp:post_modified_gmt><![CDATA[2025-03-24 12:08:44]]></wp:post_modified_gmt>
		<wp:post_name><![CDATA[test-images-with-upload]]></wp:post_name>
		<wp:status><![CDATA[publish]]></wp:status>
		<wp:post_type><![CDATA[post]]></wp:post_type>
</item>
```

Combine both the items to one standard XML, WXR format and try to import the file.

You would notice that, the media has been imported correctly, post have been imported correctly, but when you open post in edit mode, the media would look broken because, the extension to the media is not present, hence `$this->url_remap`, mapped the URLs without extension.

https://github.com/WordPress/wordpress-importer/blob/2ed3a298df5ec341eee22aadb5e2ab74435b1da9/src/class-wp-import.php#L1042-L1048

The above source code is creating the issue, because the source URL does not have any extension, hence anytime this array uses the key, it replaces the old value to new value, and new value as per the given link is removing the file extension itself.

This is the only case with an external file without extension, So we should have a condition for checking if the source URL, wp_attachment_url does not contain extension, we should not run the above lines of code to make everything works as expected.

Here is the attached video, that would demonstrate the difference between external URL and WordPress specific URL media and post import with specific scenarios.

https://github.com/user-attachments/assets/f7d19c88-8b7c-4dd0-9be2-7ac1021536d5

AswiniSudhaKattamuri:
### Summary
Prevents incorrect URL remapping when importing externally hosted media files that do not include a file extension.

### Details
Adds a defensive check before remapping resized image URLs to ensure both source and destination URLs contain a valid extension.

### Related Issue
Fixes #186

### DIFF
--- a/src/class-wp-import.php
+++ b/src/class-wp-import.php
@@ -9,6 +9,7 @@
 use WordPress\DataLiberation\URL\WPURL;
 use function WordPress\DataLiberation\URL\wp_rewrite_urls;
 
+   
 /**
  * WordPress importer class.
  */
@@ -1356,21 +1357,22 @@ class WP_Import extends WP_Importer {
 		$post_id = wp_insert_attachment( $post, $upload['file'] );
 		wp_update_attachment_metadata( $post_id, wp_generate_attachment_metadata( $post_id, $upload['file'] ) );
 
-		// remap resized image URLs, works by stripping the extension and remapping the URL stub.
-	if ( preg_match( '!^image/!', $info['type'] ) ) {
-    $parts     = pathinfo( $url );
-    $parts_new = pathinfo( $upload['url'] );
+			// remap resized image URLs, works by stripping the extension and remapping the URL stub.
+		if ( preg_match( '!^image/!', $info['type'] ) ) {
+			$parts     = pathinfo( $url );
+			$parts_new = pathinfo( $upload['url'] );
 
-    if ( empty( $parts['extension'] ) || empty( $parts_new['extension'] ) ) {
-        return $post_id;
-    }
+			if ( empty( $parts['extension'] ) || empty( $parts_new['extension'] ) ) {
+				return $post_id;
+			}
 
-    $name     = basename( $parts['basename'], ".{$parts['extension']}" );
-    $name_new = basename( $parts_new['basename'], ".{$parts_new['extension']}" );
+			$name     = basename( $parts['basename'], ".{$parts['extension']}" );
+			$name_new = basename( $parts_new['basename'], ".{$parts_new['extension']}" );
 
-    $this->url_remap[ $parts['dirname'] . '/' . $name ] =
-        $parts_new['dirname'] . '/' . $name_new;
-}
+			$this->url_remap[ $parts['dirname'] . '/' . $name ] =
+				$parts_new['dirname'] . '/' . $name_new;
+		}
+
 
 
 		return $post_id;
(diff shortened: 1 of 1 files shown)


AswiniSudhaKattamuri:
Thanks for the review!

The failing PHP test suites appear to be related to the CI WordPress
environment expecting built asset files (e.g. script-loader packages
and routes files), which are not modified by this change.

The update is limited to URL remapping logic and passes PHPCS and
relevant checks. Happy to adjust or add coverage if there’s a
preferred approach for handling this in tests.

PR #254, opened 2026-02-26
Fix: WordPress Importer does not remap noteId in block metadata when reassigning comment IDs

hbhalodia:
## Core Trac Ticket 
https://core.trac.wordpress.org/ticket/64724 

## Summary
Fixes https://core.trac.wordpress.org/ticket/64724 where Gutenberg block metadata containing noteId references to note-type comments are not updated with new comment IDs during WordPress import.

## Problem
When exporting a post with Gutenberg blocks that reference note-type comments via noteId metadata (e.g., ), the export correctly includes the comment with all required details. However, during import:

1. Comments are inserted with new auto-incremented IDs
2. The noteId values in block metadata remain unchanged
3. Block references become broken, pointing to non-existent or incorrect comments

## Example:

- **Exported**: "noteId":3252 (references comment ID 3252 on source site)
- **After import**: Block still shows "noteId":3252, but comment was assigned ID 5678 on destination site
- **Expected**: "noteId":5678

## XML file to test
[notes-import-bug-test-case.xml](https://github.com/user-attachments/files/25567688/notes-import-bug-test-case.xml)

## Solution
This PR implements automatic mapping and updating of noteId references during import:

1. **Track note-type comments**: Added $processed_comments property to store old → new comment ID mappings for comments with comment_type === 'note'
2. **Map comment IDs during import**: When inserting note-type comments, store the mapping between original and new comment IDs
3. **Update noteId references**: After all comments are imported, scan the post content and replace old noteId values with new comment IDs using simple string replacement

## Testing
✅ Imports posts with noteId metadata correctly
✅ Only processes note-type comments (ignores regular comments)
✅ Handles multiple noteId references in a single post
✅ Works with all Gutenberg blocks that support metadata
✅ No performance impact on posts without note-type comments

## Backwards Compatibility
- Fully backwards compatible
- No database schema changes
- No breaking changes to existing functionality

## AI Usage

1. Created description and code using GitHub Copilot using Claude Sonnet 4.5.
2. Reviewed the generated code and update as per need.


src/class-wp-import.php
```diff
@@ -1139,6 +1145,52 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+		}
+
+		$content = $post->post_content;
+		$updated = false;
+
+		foreach ( $this->processed_comments as $old_note_id => $new_note_id ) {
+			$search  = '"noteId":' . $old_note_id;
+			$replace = '"noteId":' . $new_note_id;
+
+			$new_content = str_replace( $search, $replace, $content );
```
westonruter: I think something more robust should be used here. Namely, this replacement should be restricted to the block comments and not done anywhere else. I think this would be a perfect use for the new streaming block parser in 6.9: https://make.wordpress.org/core/2025/11/19/introducing-the-streaming-block-parser-in-wordpress-6-9/

Otherwise (or in addition), this would be a great use for the HTML Tag Processor.

hbhalodia: Yes make sense. Will work on the above feedback.

hbhalodia: Hi @westonruter, Before implementing this using `WP_Block_Processor`, I have gone through the entire documentation and file, and seems like currently it mostly support the reading, but no as such direct modifications.

We can do some kind of modification using `extract_full_block_and_advance` using `next_block`, but it still needs to be traversed until we get all the inner blocks. It then kind of behave same as parse_blocks, because we need to traverse through all the innerblocks, check for noteId and update the array.

I have updated it in commit https://github.com/WordPress/wordpress-importer/pull/254/commits/039fc9b8404d07acc0e10e1905afc329cef70022 for review.

westonruter: OK, then it seems `WP_HTML_Tag_Processor` is what we'll need to use. 

I think it'll need to iterate over all tokens, stop at the comment ones, examine it it is a block comment, parse the JSON and then do the replacement if the prop is present, re-serialize the JSON to update via `$processor->set_modifiable_text()` (but making sure to use the same original JSON encoding flags to ensure `-->` and `---!>` is not present).

cc @dmsnell

src/class-wp-import.php
```diff
@@ -1139,6 +1145,83 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+		if ( ! $post ) {
+			return;
+		}
+
+		$new_content     = '';
+		$has_updates     = false;
+		$block_processor = new WP_Block_Processor( $post->post_content );
+
+		while ( $block_processor->next_block() ) {
+			$block = $block_processor->extract_full_block_and_advance();
```
dmsnell: this eager step not only doesn’t need to happen, but if we are going to end up doing it, `parse_blocks()` would probably be faster.

that being said, we should be able to take advantage of `get_span()` here and eek out some extra performance by avoiding parsing JSON attributes for blocks which definitely could not contain a note id.

```php
$next_note_id_at = strpos( $post->post_content, '"noteId"' );
if ( false === $next_node_id_at ) {
	return false;
}

$replacements    = array();
$block_processor = new WP_Block_Processor( $post->post_content );
while ( $block_processor->next_block() ) {
	$span = $block_processor->get_span();
	if ( $next_note_id_at > ( $span->start + $span->length ) ) {
		continue;
	}

	while ( $next_note_id_at < ( $span->start + $span->length ) ) {
		$next_note_id_at = strpos( $post->post_content, '"noteId"', $next_note_id_at + 1 );
	}

	$attrs = $block_processor->allocate_and_return_parsed_attributes();
	$old_note_id = $attrs['metadata']['noteId'] ?? null;
	if ( ! ( is_string( $old_note_id ) || is_int( $old_note_id ) ) || ! isset( $this->processed_comments[ $old_note_id ] ) {
		continue;
	}

	$attrs['metadata']['noteId'] = $this->processed_comments[ $old_note_id ];

	$json = json_encode( $attrs );
	$void = 'void' === $block_processor->get_delimiter_type() ? '/' : '';

	$replacements[] = new WP_HTML_Text_Replacement(
		$span->start,
		$span->length,
		"<!-- wp:{$block_processor->get_block_type()} {$json} {$void}-->"
	);
}

if ( 0 === count( $replacements ) ) {
	return;
}

$stitcher = new class ( $post->post_content ) extends WP_HTML_Tag_Processor {
	public function stitch( $updates ) {
		$this->lexical_updates = $updates;
		return $this->get_upated_html();
	}
}

wp_update_post(
	array(
		'ID'           => $post_id,
		'post_content' => $stitcher->stitch( $replacements )
	)
);
```

there are probably errors in this code, and maybe we should add a method `->set_attributes( array $new_attributes )` to the Block Processor, but this should do the job with substantially less work that we have in the PR right now.

the key is that the JSON attributes _must_ contain the string `"noteId"` if they contain the note id. it could be the case that a block has another `noteId` in another attribute, but even then, once we have parsed the JSON, which we would have had to do anyway, then if it doesn’t contain the metadata note id then we skip it.

this should have far fewer allocations, string copies, and writes, and all of the writes end up batched thanks to the Tag Processor.

hbhalodia: Hi @dmsnell, Thanks for the detailed update as I see ultimately we are using `WP_HTML_Tag_Processor` to get the updated HTML, So can't we use only `WP_HTML_Tag_Processor` and get the modified HTML like below?

I like the idea here to use newly added WP_Block_Processor as well!

Just the thing is usage of regex or alternatively can use some string operations below.

```php
$processor   = new WP_HTML_Tag_Processor( $post->post_content );
$has_updates = false;

while ( $processor->next_token() ) {
	if ( '#comment' !== $processor->get_token_type() ) {
		continue;
	}

	$comment_text = trim( $processor->get_modifiable_text() );

	if ( ! preg_match( '/^wp:([a-z][a-z0-9\/\-]*)\s+(\{.*\})\s*$/s', $comment_text, $matches ) ) {
		continue;
	}

	$attribute_json_string = $matches[2];
	$attributes            = json_decode( $attribute_json_string, true );

	if ( ! is_array( $attributes ) ) {
		continue;
	}

	if ( ! isset( $attributes['metadata']['noteId'] ) ) {
		continue;
	}

	$old_note_id = $attributes['metadata']['noteId'];
	if ( ! isset( $this->processed_comments[ $old_note_id ] ) ) {
		continue;
	}

	$attributes['metadata']['noteId'] = $this->processed_comments[ $old_note_id ];
	$updated_attribute_json           = wp_json_encode( $attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );

	if ( false === $updated_attribute_json ) {
		continue;
	}
	$updated_text = sprintf( ' wp:%s %s ', $matches[1], $updated_attribute_json );

	if ( $processor->set_modifiable_text( $updated_text ) ) {
		$has_updates = true;
	}
}

if ( $has_updates ) {
	wp_update_post(
		array(
			'ID'           => $post_id,
			'post_content' => $processor->get_updated_html(),
		)
	);
}
```

Let me know what could be the best approach here, may be the `WP_Block_Processor`?

hbhalodia: Updated with your suggestion in the commit - https://github.com/WordPress/wordpress-importer/pull/254/commits/3f1980e4161417be18576a4107514cf802ae2a34

Will update as needed.

Thanks,

src/class-wp-import.php
```diff
@@ -1139,6 +1145,96 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+	 * During import, note-type comments receive new auto-incremented IDs on the destination
+	 * site. Block metadata may reference those comment IDs via a `noteId` attribute, e.g.:
+	 *
+	 *   <!-- wp:heading {"metadata":{"noteId":3252}} -->
+	 *
+	 * This method uses WP_Block_Processor to walk every block opener in the post content,
+	 * checks whether its `metadata.noteId` matches an old comment ID recorded in
+	 * `$this->processed_comments`, and — if so — replaces it with the new ID. Updated
+	 * blocks are collected as WP_HTML_Text_Replacement objects and applied back to the
+	 * raw post content in one pass before saving.
```
dmsnell: there’s a lot of explanation of the internals of this algorithm which seem surprising to read here. is this by any chance a “summary” of what’s happening in the function?

what do you see as relevant to communicate about the function to someone who is calling it and seeing the docblock popup in their IDE?

src/class-wp-import.php
```diff
@@ -1139,6 +1145,96 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+	 *
+	 * Skips the post entirely when no `"noteId"` string is present in the content, or
+	 * when WP_Block_Processor is unavailable.
```
dmsnell: I might seem to be backtracking a bit by asking this, but the Block Processor is indeed new in WordPress 6.9. what is the minimum supported WordPress version  for this importer?

if we have to provide support going back we can lean more towards the Tag Processor, which was introduced in WordPress 6.2 though unfortunately, `get_modifiable_text()` didn’t come until WordPress 6.5.

Finally, if none of this is acceptable we can _lift_ the PCRE from the official parser, which some other code has done. It’s a gnarly pattern, but we want to try as hard as possible to avoid having one part of a system parse the posts one way while another part parses it another way.

---

Overall I don’t have a direction I’m trying to push you; I’m not familiar with the minimum version levels in this plugin — just trying to provide guidance whatever the policy or state might be.

src/class-wp-import.php
```diff
@@ -1139,6 +1145,96 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+
+			if (
+				! ( is_string( $old_note_id ) || is_int( $old_note_id ) ) ||
+				! isset( $this->processed_comments[ $old_note_id ] )
+			) {
+				continue;
+			}
+
+			$attributes['metadata']['noteId'] = $this->processed_comments[ $old_note_id ];
+			$json_string                      = wp_json_encode( $attributes );
```
dmsnell: this is lacking the appropriate escaping that you had in your code.

though in hindsight, we should avoid this entirely and call `serialize_block_attributes( $attributes )` which will perform proper JSON encoding _and_ prevent benign strings inside the attributes from escaping the HTML (for example, if an attribute value were to contain `-->` then it would break out of the comment and mess up the rest of the page).

src/class-wp-import.php
```diff
@@ -1139,6 +1145,96 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+		if ( empty( $replacements ) ) {
+			return;
+		}
+
+		$stitcher_class = new class ( $post->post_content ) extends WP_HTML_Tag_Processor {
+			public function stitch( $updates ) {
+				$this->lexical_updates = $updates;
+				return $this->get_updated_html();
+			}
+		};
```
dmsnell: I’ve been meaning to propose a helper function in WordPress to do this, but for your own informative reference, this is leaning on the way that the Tag Processor makes its internal updates.

if we simply used things like `substr()` here, then if any of the updates were shorter or longer than what they replace, then all of the successive `(start, length)` pairs would now point to the wrong spot. so basically this just makes the updates and adjusts the offset where those `start` values point.

src/class-wp-import.php
dmsnell: please remember I did not test any of this 😃
I’ve left some more detailed review including asking a bigger question, which is what the minimum supported WordPress version for this plugin. that may have a big impact on how much we can use these newer APIs.

hbhalodia: Yes thanks @dmsnell, Though I have tested the code on my end and then committed it.

> I’ve left some more detailed review including asking a bigger question, which is what the minimum supported WordPress version for this plugin. that may have a big impact on how much we can use these newer APIs.

Yes will take a look on that. Thanks for review.

src/class-wp-import.php
hbhalodia: > is this by any chance a “summary” of what’s happening in the function?

Yes it's summary of what happening inside the function.

> what do you see as relevant to communicate about the function to someone who is calling it and seeing the docblock popup in their IDE?

I understood it now, since this would not be a kind of a public function that user's will use so they do need the documentation, there is no any such need to add these much details. I will update it with short function doc comment. 🙇

src/class-wp-import.php
hbhalodia: > I might seem to be backtracking a bit by asking this, but the Block Processor is indeed new in WordPress 6.9. what is the minimum supported WordPress version for this importer?

Yeah I totally agree on this, it's a new introduction in 6.9 version. The minimum version supported in [WordPress Importer plugin ](https://wordpress.org/plugins/wordpress-importer/) is `5.2`.

But here is my question,
1. Since we have an issue with the comments with type `note`, which is introduced in WP 6.9 only. Before that there was nothing called as `notes`.
2. So even if we use this, check for `WP_Block_Processor` class available or not, then continue the process would be fair right?

I only think of a scenario is, someone used `WP 6.9` to add the notes, then `degrade the WP version`, then exported the XML file, the XML file generated would still have the comments with type note on that. So when that XML file is imported to a version that is lower than 6.9, the notes aka comment would have inserted, but it would eventually not be visible to user on edit screen, as minimum required version for notes is `6.9`.

> if we have to provide support going back we can lean more towards the Tag Processor, which was introduced in WordPress 6.2 though unfortunately, get_modifiable_text() didn’t come until WordPress 6.5.

Yes we can do, but the thing as I mentioned is specifically we are targeting the comments with type note, that was came along with `WP_Block_Processor` in 6.9.

> Finally, if none of this is acceptable we can lift the PCRE from the official parser, which some other code has done. It’s a gnarly pattern, but we want to try as hard as possible to avoid having one part of a system parse the posts one way while another part parses it another way.

Hmm, Yeah that is the end stage to work with. Regexsss 😄 

> Overall I don’t have a direction I’m trying to push you; I’m not familiar with the minimum version levels in this plugin — just trying to provide guidance whatever the policy or state might be.

Yes, thanks for the details. As per minimum version for both notes and WP_Block_Processor, I guess we can go ahead with the above IMO?

src/class-wp-import.php
hbhalodia: Thanks for the suggestion, will take a look and update.

src/class-wp-import.php
hbhalodia: > I’ve been meaning to propose a helper function in WordPress to do this, but for your own informative reference, this is leaning on the way that the Tag Processor makes its internal updates.
> 
> if we simply used things like `substr()` here, then if any of the updates were shorter or longer than what they replace, then all of the successive `(start, length)` pairs would now point to the wrong spot. so basically this just makes the updates and adjusts the offset where those `start` values point.

Okay, so helper function should be introduced in core with another trac ticket IMO.

Yes, substr could not be used as mentioned, but I see many places in the core which uses the same techniques to make the replacement and provide back the updated HTML. So we can use the same here as well?

Ref - https://github.com/WordPress/wordpress-develop/blob/b8855c4ca7d3176754d9c247c428736ee8c4ffc7/src/wp-includes/block-template.php#L350

https://github.com/WordPress/wordpress-develop/blob/b8855c4ca7d3176754d9c247c428736ee8c4ffc7/src/wp-includes/script-loader.php#L3834

dmsnell: yeah I was just noting that it’s not about the Tag Processor; rather it’s just about having a convenient way to apply all these updates

src/class-wp-import.php
dmsnell: > I only think of a scenario is, someone used WP 6.9 to add the notes, then degrade the WP version, then exported the XML file, the XML file generated would still have the comments with type note on that. So when that XML file is imported to a version that is lower than 6.9, the notes aka comment would have inserted, but it would eventually not be visible to user on edit screen, as minimum required version for notes is 6.9.

You raise a good scenario, but didn’t write it out. I think it means we _should_ try to make sure this works pre-6.9.

 1. I export my 6.9 site containing notes.
 2. I import the WXR into an old site.
 3. Eventually I update the old site.

If we skip remapping the note ids then we’ll not only have lost the ability to connect those notes whenever the site is updated, but we’ll be actively linking posts to the wrong notes. Isn’t that more or less right?

> Hmm, Yeah that is the end stage to work with. Regexsss 😄

In this case it’s not too bad. The PCRE in `class-wp-block-parser.php` is well-vetted and appropriate for this purpose. In early explorations for the `WP_Block_Parser` class I looked at using that, but the ability to remove the dependency on the `pcre` extension was too valuable.

It might be quite unexpected, but block posts are defined at this level above HTML. We could be breaking HTML in rare cases, but that is the spec, for a number of historic reasons.

It should be possible to reuse the block parser without extracting the PCRE bits, though it might look a bit odd, and with this method we can’t avoid eagerly parsing JSON, but we get some of the perks without the extra duplication and maintenance.

```php
$parser = new WP_Block_Parser();
$parser->document = $post->post_content;
do {
	$next_token = $this->next_token();
	list( $token_type, $block_name, $attrs, $start_offset, $token_length ) = $next_token;

	if ( 'block-opener' !== $token_type && 'void-block' !== $token_type ) {
		continue;
	}

	…
} while ( 'no-more-tokens' !== $token_type );
```

Of course, pulling the PCRE out would allow the optimization to skip JSON parsing, but 🤷‍♂️ tradeoffs...

src/class-wp-import.php
```diff
@@ -1139,6 +1145,82 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+
+		wp_update_post(
+			array(
```
westonruter: Very important: `wp_update_post()` expects pre-slashed content, when an array is passed: https://github.com/WordPress/wordpress-develop/blob/de3aa66eae68de88b0ce543aa69941f81e2b92e2/src/wp-includes/post.php#L5255-L5267

The supplied content will get unslashed by `wp_post()`:

https://github.com/WordPress/wordpress-develop/blob/de3aa66eae68de88b0ce543aa69941f81e2b92e2/src/wp-includes/post.php#L4920

By casting to an object, the slashing is applied by `wp_update_post()`.

```suggestion
			// Cast to object to ensure wp_update_post() will add the required slashes.
			(object) array(
```

westonruter: Without this, intentional slashes (e.g. in JSON) would get corrupted.

hbhalodia: Thanks @westonruter, Will update it on the final version we go ahead with.

src/class-wp-import.php
hbhalodia: > If we skip remapping the note ids then we’ll not only have lost the ability to connect those notes whenever the site is updated, but we’ll be actively linking posts to the wrong notes. Isn’t that more or less right?

Yeah makes sense, I did'nt think of user updating the old site to 6.9 after import with notes on older site.

> It should be possible to reuse the block parser without extracting the PCRE bits, though it might look a bit odd, and with this method we can’t avoid eagerly parsing JSON, but we get some of the perks without the extra duplication and maintenance.

In that case it's good to use tag_processor as mentioned here in comment https://github.com/WordPress/wordpress-importer/pull/254#discussion_r2862666429, It does use simple regex match for the comment delimeter to catch for the attribute and update the same? Then again the question arises, upto how older version of WP we should allow this, as set_modifiable_text is added in 6.7 version.

Also, I did'nt quite get the example you shared. We do not have any such way to update the text using Block_Parser. If I am thinking it correctly, we would find the $start_offset, $token_length, create the replacements array as we have currently and update it using tag_processor? the thing that is currently added in the PR? but use WP_Block_Parser?

dmsnell: > In that case it's good to use tag_processor
> upto how older version of WP we should allow this

🤷‍♂️ it seems more-or-less equally hard or equally easy to cover all the way in support of what this plugin requires as a minimal version: WordPress 5.2. one thing I like to do is leave factual notes on old code communicating _when exactly_ it can be replaced.

```php
// Replace with Tag Processor once minimal supported WordPress version is 6.7.
```

this is a unique case because usually we’re making a tradeoff on correctness whether to use a newer API or not. here, it’s mostly a matter of convenience  from the code perspective. that gives us more freedom I think.

if you review my work I’m sure you’ll find that I missed this or messed up in multiple places, but I don’t think we should say “this requires WordPress 5.2” and then ship degraded functionality on that version.

> We do not have any such way to update the text using Block_Parser. If I am thinking it correctly, we would find the $start_offset, $token_length, create the replacements array as we have currently and update it using tag_processor? the thing that is currently added in the PR? but use WP_Block_Parser?

Yes, exactly. So we have two options. `WP_Block_Parser->next_token()` returns those starting byte offsets and byte lengths for the entire block delimiter, and we know this fact: the JSON attributes start at the first `{` within that range and end at the last `}` within that range.

We can use `strcspn()` to find the starting `{` and `strrpos()` to find the last one, then we know what text to replace in the original document.

Alternatively, if we directly lift out the PCRE from `WP_Block_Parser` we can use the returned ranges of the JSON object. That duplicates code, which is more work to maintain, but the PCRE there is “battle-tested” and unlikely to change quickly. More so, it matches the behavior in Core.

The primary tradeoff here is that:
 - if we use `next_token()` we are eagerly parsing the JSON but we don’t have to maintain the code.
 - if we lift out the PCRE we will get full control but it will be harder to reason about and maintain

westonruter: I don't think WP 5.2 is something to worry about here. Notes were only introduced in 6.9. If someone is trying to import content from WP 6.9 into a WP 5.2 site, they're not going to have notes anyway.

dmsnell: sure @westonruter, _but_ maybe we should consider updating the minimum version before we start building on newer versions?

also, we do have the use-case of people importing into an old site, and if we bail on remapping, then we’re doing more harm than wiping them out. once that site does update, it will have corrupted note links.

🤷‍♂️

hbhalodia: Thanks for the detail @dmsnell @westonruter, I have updated the PR to use with first option, that is, use of WP_Block_Parser and update the attributes, then use string functions to replace the original content with new one.

The only thing as of worth noticing is use of `serialize_block_attributes` function, as we are having a lot of discussion around the minimum supported version of this plugin. Since the mentioned function was introduced in version `5.3.1` and supported version of this plugin is `5.2`. I just wanted to flag it, but we can literally go ahead with this, would not be having impact of larger users.

src/class-wp-import.php
hbhalodia: @dmsnell, I have updated it, but just to flag this was introduced in 5.3.1 and currently minimum supported version is 5.2.

Details added more here -- https://github.com/WordPress/wordpress-importer/pull/254#discussion_r2882023019

src/class-wp-import.php
hbhalodia: There is no need to add this right now, as we have used string function and WP_Block_Parser.

src/class-wp-import.php
hbhalodia: This is now added to the commit -- https://github.com/WordPress/wordpress-importer/pull/254/commits/8db9ccff672e9c053c40665bd1c7175a904171c6

src/class-wp-import.php
```diff
@@ -1139,6 +1145,95 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+				continue;
+			}
+
+			$json_start  = $start_offset + $attribute_json_start;
+			$json_length = $attribute_json_end - $attribute_json_start + 1;
+
+			$attrs['metadata']['noteId'] = $this->processed_comments[ $old_note_id ];
+			$replacements[]              = array( $json_start, $json_length, serialize_block_attributes( $attrs ) );
+
+		} while ( 'no-more-tokens' !== $token_type );
```
dmsnell: this is convenient for continuing, but if there are no block comment delimiters within the post we will be running all of this on broken data.

it might be wise to make it a bit more verbose with the `while() {}` form and an early-abort at the top of the loop.

```php
while ( $parser->offset < $end ) {
	list( … ) = $parser->next_token();
	if ( 'no-more-tokens' === $token_type ) {
		break;
	}

	…
}
```

src/class-wp-import.php
```diff
@@ -1139,6 +1145,95 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+
+		// Apply replacements in reverse order to avoid affecting offsets of later replacements.
+		$replacements     = array_reverse( $replacements );
+		$updated_content  = $post->post_content;
+
+		// Loop through each replacement and update the content string with the new JSON attributes.
+		foreach ( $replacements as $replacement ) {
+			list( $offset, $length, $new_json ) = $replacement;
+			$updated_content = substr_replace( $updated_content, $new_json, $offset, $length );
+		}
```
dmsnell, line 1227: keep in mind that this could have heavy memory allocation requirements. with the same information, we could build a new string and only incur the kind of allocations we expect when growing a string. this is a builder approach rather than a replacement approach.

```php
$updated_content = '';
$was_at          = 0;
foreach ( $replacements as $replacement ) {
	list( $offset, $length, $new_json ) = $replacement;
	
	$pre_length       = $offset - $was_at;
	$updated_content .= substr( $post_content, $was_at, $pre_length ) . $new_json;
	$was_at           = $offset + $length;
}

$updated_content .= substr( $post_content, $was_at );
```

these will likely be similar in practice, but when replacing backwards, we have to start by fully-allocating the entire string on every replacement, whereas with the builder approach, we likely have a much smaller count, and one related to the overall length of the string.

I’d be curious to see some real data here, and what you have already is fine until we have that data, but I do think the worst-case scenarios might benefit greatly from the reduced-allocation and reduced-copy version.

dmsnell (commented):
@hbhalodia what do you need still from anyone on this ticket? I’ve been away for a few days and am trying to catch up and want to make sure this isn’t overlooked.

src/class-wp-import.php
hbhalodia: Thanks @dmsnell for review, Updated the same.

src/class-wp-import.php
hbhalodia: Thanks updated the same. I do have tested with the file shared in the original ticket -- https://core.trac.wordpress.org/ticket/64724

hbhalodia:
> @hbhalodia what do you need still from anyone on this ticket? I’ve been away for a few days and am trying to catch up and want to make sure this isn’t overlooked.

Just wanted the review and best practices and we can land this PR if everything is working as expected.

src/class-wp-import.php
```diff
@@ -1139,6 +1145,99 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+		if ( empty( $this->processed_comments ) ) {
+			return;
+		}
+
+		$post = get_post( $post_id );
+		if ( ! $post ) {
+			return;
+		}
+
+		$is_contain_notes = strpos( $post->post_content, '"noteId"' );
```
dmsnell: `str_contains()` is polyfilled, so it should be fine to use this on all currently-supported versions of WordPress.

```php
if ( ! str_contains( $post->post_content, '"noteId"' ) ) {
	return;
}
```

src/class-wp-import.php
```diff
@@ -1139,6 +1145,99 @@ protected function process_post_comment_metas( $comment_id, $commentmeta ) {
...
+				! isset( $this->processed_comments[ $old_note_id ] )
+			) {
+				continue;
+			}
+
+			$attribute_string     = substr( $post->post_content, $start_offset, $token_length );
+			$attribute_json_start = strcspn( $attribute_string, '{' );
+			$attribute_json_end   = strrpos( $attribute_string, '}' );
+
+			if ( false === $attribute_json_end || $attribute_json_start >= $attribute_json_end ) {
```
dmsnell, line 1202: if the parser returned these attributes then we shouldn’t need the bounds-checking. it’s already been performed.

if we leave it in for the sake of explicitness, then there’s an error where the starting bound could be `false` and it will fail to abort.

so I think it might be appropriate to either remove this step entirely (it should never trigger here) or add `false === $attribute_json_start` to the conditions.

dmsnell (approved):
this all looks good to me. it seems like this was a great hassle and you were awesome going back and forth on it with me and with others. the end product is something that will work on any WordPress install regardless of the (supported) version and I think that’s great.

I noticed there are no added tests, and I didn’t look at the state of tests in this repo. I’m assuming you have tested this out? I sure haven’t. it would be nice if we (a) knew this actually works, and (b) included some assertions to demonstrate it working and to prevent regressions in the future.

src/class-wp-import.php
hbhalodia: Done Thanks,

src/class-wp-import.php
hbhalodia: Done Thanks

hbhalodia:
> I noticed there are no added tests, and I didn’t look at the state of tests in this repo. I’m assuming you have tested this out? I sure haven’t. it would be nice if we (a) knew this actually works, and (b) included some assertions to demonstrate it working and to prevent regressions in the future.

Yeah I have not addded tests cases for now. I will test this out with different post types, blocks etc  and will add detailed XML files here for others to test, and would also add the test cases once I spin up test enviornment for this.

Thanks,

hbhalodia:
Hi @dmsnell, I have added a normal unit test case which checks for the string `{"noteId":<note-number>}` with the new mapped ID in the content. That was a quick and easy thing we can check if our new content has newely mapped ID.

Also added the XML file agains to check the unit test for.

Thanks,

src/class-wp-import.php
dmsnell: only when I linked to this as an example for another PR did I notice that `$attribute_json_start` can never be `false`, only past the length of the `$attribute_string`.

so you had it right before and that was a review blunder on my part. it now makes more sense why the `$start >= $end` check was there. sorry!

hbhalodia: Thanks, Have reverted the same. 

Probably I could have thought of more of that case and should have replied before addressing that!

hbhalodia:
Hi @dmsnell, Sorry to followup again on this. 

Can this be merged and can we release the plugin? or do we need to have more work on this?

Thanks,

dmsnell:
@hbhalodia do you need someone to merge this? sorry if I misunderstood; I did not think you were waiting on me.

hbhalodia:
> @hbhalodia do you need someone to merge this? sorry if I misunderstood; I did not think you were waiting on me.

Yes, I do not have enough permissions to merge this. Also, If someone can merge and we can release the plugin if everything looks good?

westonruter:
It looks like @adamziel is most recently active on this repo. Maybe he would feel comfortable merging? There CI issues that need to be investigated, however.

dmsnell:
it seems like this is a clear failure of the test runners, and sadly I have little time available to dedicate here.

I think one potential fix would be to create another branch, run `rm -rf composer.lock && composer install` and see if we get a clean install.

with the API docs work I saw similar issues where `composer` obstructed installation for security issues in the dependencies. there’s a workaround, and that might be acceptable here, but it’d be nice if we can resolve this just by updating the dependencies — might take some work in `composer.json` to update.

hopefully the test suite is comprehensive enough to catch any breakages due to the updates, and I think it’s probably mostly or all for the test suite itself, so that might be an easy process.

@hbhalodia if you wait on me it may take a while. I don’t expect you to go through the effort, but a good way to speed up this process is probably to create a new PR updating the now-out-of-date dependencies. if `composer install` runs successfully we can rebase this branch or merge the updates into it and that should fix it. I’d rather see us fix the tests first and then keep this separate than squish the two changes together.

PR #259, opened 2026-04-29
Fix: Update composer.json to run the test suites

hbhalodia:
- Update composer.json to add phpunit version 8.5 and 9.
- Add `doctrine/instantiator` to instantiate objects. (Resolved unit test with php 8.2 WP latest and php 8.2 wp 6.2)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "wp-coding-standards/wpcs": "^3.1",
         "phpcompatibility/phpcompatibility-wp": "^2.1",
         "yoast/phpunit-polyfills": "^1.0.1",
-        "phpunit/phpunit": "^5.7.21 || ^6.5 || ^7.5"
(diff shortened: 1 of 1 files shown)


hbhalodia:
Hi @dmsnell, Have updated the composer dependencies and its working as expected. The only tests are failing because of compatibility issue for PHPUnit version and WP compatibility.

I have tried multiple things in PR - https://github.com/WordPress/wordpress-importer/pull/258, But was unable to resolve the test failures, but yes have resolved the issue with composer.json which were not installed as per the comment - https://github.com/WordPress/wordpress-importer/pull/254#issuecomment-4298017389

Let me know if we can move forward with this, merge and rebase the PR - https://github.com/WordPress/wordpress-importer/pull/254?

Or to fix the test, what can be done to resolve unit test with less than WP 5.7 version added in tests.yml? should we skip those, remove or is there any alternatives?

PR #264, opened 2026-08-26
Don't import the _wp_font_face_file meta key

mcliwanow:
`is_valid_meta_key()` skips `_wp_attached_file` and `_wp_attachment_metadata` because the importer regenerates both from the attachment it downloads. WordPress 6.5 added a third key of the same kind, `_wp_font_face_file`, and the list never picked it up.

Font faces are not attachments, so the importer never fetches the file one names. The imported value points at a file the destination site does not have, and nothing regenerates it. Everything that makes a font face work is in `post_content`, so the font face itself still imports unchanged.

### CI

`master` has been red on every scheduled run since at least July, for reasons unrelated to each other or to the change above. Fixed here so this can be seen to pass:

- PHPCS installed PHPUnit it does not use, and every version `composer.json` pins is blocked by CVE-2026-24765, so the install failed before any style check ran. It now drops the requirement first, the same way `tests.yml` already does for WP 5.9 and up.
- WP below 5.9 predates the Polyfills bootstrap in Core and needs PHPUnit 5 - 7, which are end of life with no release outside that advisory. It is allowed for those builds only, so `composer.json` is untouched and other advisories still block.
- PHP 8.2 was the only version installing with `--ignore-platform-req=php`, added when 8.2 was an RC. Ignoring the platform check now resolves packages which require PHP 8.3, so the run died on syntax it could not parse. It installs the same way 8.1 and 8.3 do.
- WP 7.0 raised the minimum to PHP 7.4, so pairing 7.2 and 7.3 with `latest` failed before any test ran. They build against WP 6.8 now, the highest release they can run.


zaerl (approved):
It's ok. Thanks for the help.

composer.json
```diff
@@ -28,6 +28,13 @@
             "dealerdirect/phpcodesniffer-composer-installer": true
         },
         "process-timeout": 900,
-        "lock": false
+        "lock": false,
+        "policy": {
+            "advisories": {
+                "ignore-id": [
+                    "PKSA-z3gr-8qht-p93v"
```
mcliwanow: reviewing if this is the best fix we can do

